PR #4, opened 2022-09-07
set up package structure so actual work can begin

skhugh:
Sets up package structure so actual work can begin.
Empty classes are created so whole structure can be committed.

Fixes #

### Checklist

- [x] Didn't break anything


hackerwins (commented):
LGTM 👍

PR #6, opened 2022-09-14
Implement some basis types needed for other types and data structures.

skhugh:
- Implements CrdtElement, ActorID and TimeTicket classes which are needed for other types and data structures such as RGATreeList and RHT.
- Changes names for acronyms to conform to following Kotlin conventions:

> When using an acronym as part of a declaration name, capitalize it if it consists of two letters (IOStream); capitalize only the first letter if it is longer (XmlFormatter, HttpInputStream).

Fixes #

### Checklist

- [x] Didn't break anything


hackerwins (approved):
LGTM. 👍

yorkie/src/main/kotlin/dev/yorkie/document/time/ActorID.kt
```diff
@@ -0,0 +1,25 @@
+package dev.yorkie.document.time
+
+/**
+ * [ActorID] is used to identify who is making changes to the document.
+ * It is hexadecimal string and should be generated by unique value.
+ */
+@JvmInline
+internal value class ActorID(val id: String) : Comparable<ActorID> {
```
hackerwins, line 8: Initials are expressed as Uppercase in Go. Do we have to express `ActorID` as `ActorId` in Kotlin?

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

skhugh: [Kotlin doc states](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names) that we should use all uppercase for 2-letter acronyms and camel case for more letters.
So it seems `ActorID` is the better choice.

daeyounglnc: According to [Kotlin doc states](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names), it seems that both `CrdtArray` and `CrdtElement` classes should be changed to `CRDT~`, isn't it?

hackerwins: I saw the examples from the link left by @skhugh.

> When using an acronym as part of a declaration name, capitalize it if it consists of two letters (IOStream); capitalize only the first letter if it is longer (XmlFormatter, HttpInputStream).

PR #7, opened 2022-09-15
Implements SplayTree

skhugh:
Implements `SplayTree` which is needed for `RGATreeList` and `RGATreeSplit`.

This is very similar to `yorkie-js-sdk` version of `SplayTree` with an attempt to make `SplayTreeNode` non-visible to call-site, which is more familiar with JAVA/Kotlin developers.

To make this possible, it is implemented as a set.
This way, `RGATree[List|Split]Nodes` don't need to extend `SplayTreeNode` and since each node will be identified with a unique TimeTicket, implementing as a set should be okay.

Keep in mind that it is an *attempt* and if it seems implausible to implement `RGATreeList` and `RGATreeSplit` with this version, we are always open to switch back to `yorkie-js-sdk` version.

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/test/kotlin/dev/yorkie/util/SplayTreeSetTest.kt
```diff
@@ -0,0 +1,114 @@
...
+    fun setUp() {
+        target = SplayTreeSet { it.length }
+    }
+
+    @Test
+    fun `should conform to specs when insertions and splaying occur`() {
+        assertEquals(0, target.length)
+
+        target.insert("A2")
+        assertEquals("[2,2]A2", target.getStructureAsString())
```
hackerwins, line 20: I think `getStructureAsString` is intuitive and good. 👍 
It would be good to change other SDKs as well.

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
```diff
@@ -0,0 +1,364 @@
+package dev.yorkie.util
+
+import com.google.common.annotations.VisibleForTesting
```
hackerwins, line 3: This is a cool feature. There were cases where other SDKs exposed privates to the public because of test code.

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
```diff
@@ -0,0 +1,364 @@
...
+        }
+        if (target > node.length) {
+            error("out of index range: pos: $target > node.length: ${node.length}")
+        }
+        return ValueToOffset(node.value, target)
+    }
+
+    /**
+     * finds the index of the given value in BST
+     */
```
hackerwins, line 66: Are there any recommended comment styles in Kotlin? 🤔 I think we need to consider the style when writing reference documentation in the future. It would be nice to add a lint rule for comments if possible.

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
```diff
@@ -0,0 +1,364 @@
...
+
+    /**
+     * returns the Value and offset of the given index.
+     */
+    @Suppress("UNCHECKED_CAST")
+    fun find(pos: Int): ValueToOffset<V> {
+        var target = pos
+        val root = root
+        if (root == null || target < 0) {
+            return ValueToOffset.Empty as ValueToOffset<V>
```
hackerwins, line 44: Do we need to add UNCHECKED_CAST because of `as`?

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
```diff
@@ -0,0 +1,364 @@
...
+            } else if (node.right != null && node.leftWeight + node.length < target) {
+                target -= node.leftWeight + node.length
+                node = requireNotNull(node.right)
+            } else {
+                target -= node.leftWeight
+                break
+            }
+        }
+        if (target > node.length) {
+            error("out of index range: pos: $target > node.length: ${node.length}")
```
hackerwins, line 59: I wonder how Android-based SDKs handle errors.

1. Use `error` not defining things like custom errors?
2. How do we receive and handle this error on the caller?

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
```diff
@@ -0,0 +1,364 @@
...
+
+import com.google.common.annotations.VisibleForTesting
+
+/**
+ * SplayTreeSet is weighted binary search tree set based on Splay tree.
+ * original paper on Splay Trees:
+ * @link https://www.cs.cmu.edu/~sleator/papers/self-adjusting.pdf
+ */
+// should SplayTree implement MutableSet?
+internal class SplayTreeSet<V>(private val lengthCalculator: (V) -> Int = { 1 }) {
```
hackerwins, line 11: I think that the idea, of dealing with creation-time-based sets is novel. 😃
Are there any constraints on becoming a TreeSet's Value?

hackerwins (commented):
Thanks for your contribution.

I was curious about the various features of Kotlin, so I left a few questions.

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
skhugh: Yes, this comment style follows the convention.
https://kotlinlang.org/docs/coding-conventions.html#documentation-comments

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
skhugh: this annotation is very convenient, but doc states that it is not a good practice for public and protected visibility. :(
> Do not use this interface for public or protected declarations: it is a fig leaf for bad design, and it does not prevent anyone from using the declaration---and experience has shown that they will.

but since it is used on an internal class here, we are good to go :)

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
skhugh: 1. We can use custom errors or exceptions provided by Java & Kotlin.

in fact, `error` is a helper method provided by Kotlin which throws `IllegalStateException`.
Which I thought would be a good choice at the time, but I think it should be changed to throw `IndexOutOfBoundsException` :)

2. we can simply `try-catch`. But Kotlin does not have checked exceptions (https://kotlinlang.org/docs/exceptions.html#checked-exceptions)

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
skhugh: Yes although it seems ugly, this pattern is used quite a lot for optimizations for empty or default types.
This way we can share a single instance across any generic types.

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
skhugh: Only constraint that comes to mind is that it doesn't allow duplicates.
Therefore if the caller wants use it with duplicated primitive values they need to create a wrapper which is a little memory-inefficient.
But since we are planning to use this for objects with a unique TimeTicket, I think we have no problem :)

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
```diff
@@ -0,0 +1,364 @@
+package dev.yorkie.util
+
+import com.google.common.annotations.VisibleForTesting
+
+/**
+ * SplayTreeSet is weighted binary search tree set based on Splay tree.
+ * original paper on Splay Trees:
+ * @link https://www.cs.cmu.edu/~sleator/papers/self-adjusting.pdf
+ */
+// should SplayTree implement MutableSet?
```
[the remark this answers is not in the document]
skhugh: done 5471cd4

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
```diff
@@ -0,0 +1,364 @@
...
+    private val valueToNodes = mutableMapOf<V, Node<V>>()
+
+    @VisibleForTesting
+    var root: Node<V>? = null
+        private set
+
+    val length
+        get() = root?.weight ?: 0
+
+    private val maximum: Node<V>?
```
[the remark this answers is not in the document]
skhugh: done 5471cd4

yorkie/src/test/kotlin/dev/yorkie/util/SplayTreeSetTest.kt
```diff
@@ -0,0 +1,114 @@
...
+        val nodes = buildSampleTree()
+        target.deleteRange(nodes[2], nodes[7])
+        assertEquals(nodes[7], target.root?.value)
+        assertEquals(nodes[2], target.root?.left?.value)
+        assertEquals(9, target.root?.weight)
+        assertEquals(6, target.root?.left?.weight)
+        assertIfRangeRemoved(nodes, 3..6)
+    }
+
+    // FIXME: need clarification for test name
```
hackerwins, line 81:
```suggestion
    // FIXME(skhugh): need clarification for test name
```

yorkie/src/test/kotlin/dev/yorkie/util/SplayTreeSetTest.kt
```diff
@@ -0,0 +1,114 @@
...
+    @Test
+    fun `should delete nodes in range when deleteRange is invoked with no right boundary`() {
+        val nodes = buildSampleTree()
+        target.deleteRange(nodes[6])
+        assertEquals(nodes[6], target.root?.value)
+        assertEquals(22, target.root?.weight)
+        assertIfRangeRemoved(nodes, 7..8)
+    }
+
+    // FIXME: need clarification for test name
```
hackerwins, line 69:
```suggestion
    // FIXME(skhugh): need clarification for test name
```

hackerwins (approved):
Thanks for your contribution.
I was able to learn about Kotlin by reviewing the PR.

PR #12, opened 2022-10-04
Implements CrdtArray

7hong13:
Implements `CrdtArray` which will be used in ``.

It is implemented based on `CRDTArray` in `yorkie-js-sdk`.
I marked `NOTE` on comments which are different from original ones.

Things to know:
- `RgaTreeListNode` is currently visible from outside. This architecture is to be refactored later on. (will take benchmark tests and examine its effect on performance)
- Functions such as `toJson` will be renamed later not to expose its domain directly.
- Some abstract functions in `CrdtElement` will be implemented when they are actually needed.
- some function names has been changed based on the following discussion:
 https://github.com/yorkie-team/yorkie-ios-sdk/pull/13#discussion_r987477101
(release/purge -> delete, delete -> remove)

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -1,3 +1,162 @@
...
+     * Returns the sub path of the given [createdAt] element.
+     */
+    override fun subPathOf(createdAt: TimeTicket): String? {
+        return elements.subPathOf(createdAt)
+    }
+
+    // NOTE(7hong13): Original comment from JS SDK is as follows:
+    // `purge` physically purge child element.
+    /**
+     * Physically purges [elements].
```
hackerwins, line 31:
```suggestion
     * Physically purges the given [element].
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -1,3 +1,162 @@
...
+
+    /**
+     * Deletes the element of given [index] and [editedAt].
+     */
+    fun deleteByIndex(index: Int, editedAt: TimeTicket): CrdtElement? {
+        return elements.deleteByIndex(index, editedAt)
+    }
+
+    /**
+     * Traverse the descendants of this array.
```
hackerwins, line 96:
```suggestion
     * Traverses the descendants of this array.
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -1,3 +1,162 @@
...
+    override fun toJson(): String {
+        val json = mutableListOf<String>()
+        return forEach { json.add(it.toJson()) }.run {
+            json.joinToString(",")
+        }
+    }
+
+    // TODO(7hong13): implement toJS method (json parser)
+    /**
+     * Returns the javascript object of this array.
```
hackerwins, line 121:
```suggestion
     * Returns the JavaScript object of this array.
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtContainer.kt
```diff
@@ -0,0 +1,16 @@
+package dev.yorkie.document.crdt
+
+import dev.yorkie.document.time.TimeTicket
+
+/**
+ * [CrdtContainer] represents CRDTArray or CRDtObject.
```
hackerwins, line 6:
```suggestion
 * [CrdtContainer] represents CrdtArray or CrdtObject.
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/RgaTreeList.kt
```diff
@@ -92,7 +91,7 @@ internal class RgaTreeList {
     }
 
     /**
-     * Physically purges element.
+     * Physically purges [element].
```
hackerwins, line 94:
```suggestion
     * Physically purges the given [element].
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/RgaTreeList.kt
```diff
@@ -179,13 +178,21 @@ internal class RgaTreeList {
...
-    fun getNodesInListExceptHead(): List<RgaTreeListNode> {
-        return buildList {
-            var curr = dummyHead.next
-            while (curr != null) {
-                add(curr)
-                curr = curr.next
+    // NOTE(7hong13): Original comment from JS SDK is as follows:
+    // `getLastCreatedAt` returns the creation time of last elements.
+    /**
+     * Returns the creation time of last element.
```
hackerwins, line 184:
```suggestion
     * Returns the creation time of the last element.
```

hackerwins (commented):
Overall it looks good.
I leave a few comments.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -1,3 +1,162 @@
...
+     * Moves the given [createdAt] element after the [prevCreatedAt] element.
+     */
+    fun moveAfter(prevCreatedAt: TimeTicket, createdAt: TimeTicket, executedAt: TimeTicket) {
+        elements.moveAfter(prevCreatedAt, createdAt, executedAt)
+    }
+
+    /**
+     * Returns the element of the given [createdAt].
+     */
+    fun get(createdAt: TimeTicket): CrdtElement? {
```
skhugh, line 56: maybe we could make this `operator fun`

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -1,3 +1,162 @@
...
+        val node = elements.get(createdAt)
+        if (node?.isRemoved == true) return null
+        return node
+    }
+
+    /**
+     * Returns the element of the given [index].
+     */
+    fun getByIndex(index: Int): CrdtElement? {
```
skhugh, line 65: maybe we could make this `operator fun`

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtElement.kt
```diff
@@ -10,6 +10,12 @@ internal abstract class CrdtElement(
     movedAt: TimeTicket? = null,
     removedAt: TimeTicket? = null,
 ) {
+    abstract fun toJson(): String
```
skhugh, line 13: I think it is better to delay implementing `toJson`, `toSortedJson`, `deepCopy` until they are actually needed.
I'm hoping we could implement these functions easily with JSON serializer/deserializer libraries such as GSON.

hackerwins: @skhugh Interesting. How can we use GSON?

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
7hong13: @hackerwins we can use `operator fun` in kotlin for some benefits, but then we need to rename this function to `get`, instead of `getByIndex`.
is it ok to have different function name with js-sdk?

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtElement.kt
daeyounglnc: @hackerwins Please check this page : https://github.com/google/gson
Just add gson dependency to gradle, and use it according to guide docs.
It's the most popular JSON processing library for Java/Kotlin users.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -1,3 +1,162 @@
...
+     * Returns the JSON encoding of this array.
+     */
+    override fun toJson(): String {
+        val json = mutableListOf<String>()
+        return forEach { json.add(it.toJson()) }.run {
+            json.joinToString(",")
+        }
+    }
+
+    // TODO(7hong13): implement toJS method (json parser)
```
daeyounglnc, line 119: Currently, we use the comment like `TODO($name)` when developer need to implement the function body. 
But, I think, we would use `TODO(String)` function would rather than `TODO($name)` comment for more formal way. Because it can throw the Exception if it wasn't implemented.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
hackerwins: Never know about `operator fun` before. Can we use `operator fun` for `TimeTicket` and `Index` at the same time(method overloading)?

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtElement.kt
hackerwins: Thanks for the introduction. I'd also like to know how to use GSON when creating JSON from Yorkie Document. 🤔

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
7hong13: @hackerwins yes, it seems like method overloading is allowed here (I see no compile time error)

yorkie/src/test/kotlin/dev/yorkie/document/crdt/CrdtArrayTest.kt
```diff
@@ -0,0 +1,132 @@
+package dev.yorkie.document.crdt
+
+import dev.yorkie.document.time.ActorID
+import dev.yorkie.document.time.TimeTicket
+import junit.framework.TestCase.assertEquals
```
skhugh, line 5: `junit.framework.TestCase` is deprecated.
We should use `assert*` functions from `org.junit.Assert` instead.

yorkie/src/main/kotlin/dev/yorkie/document/time/TimeTicket.kt
```diff
@@ -23,26 +23,6 @@ internal data class TimeTicket(
...
-    override fun hashCode(): Int {
-        var result = lamport.hashCode()
-        result = 31 * result + delimiter
-        result = 31 * result + actorID.hashCode()
-        return result
-    }
-
     companion object {
         const val INITIAL_DELIMITER = 0
         const val MAX_DELIMITER = Int.MAX_VALUE
```
hackerwins, line 28: Not directly related to this PR, but I leave the question below.

The type of delimiter is `uint32`, and the maximum value is also the maximum value of `uint32`. Is there a way to represent a `uint32` in Kotlin?

https://github.com/yorkie-team/yorkie/blob/main/pkg/document/time/ticket.go#L30

hackerwins: It's unlikely to happen, but it can cause potential problems, so I recommend using the same type. How about leaving a comment on this?

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -1,3 +1,162 @@
...
+
+    /**
+     * Returns the sub path of the given [createdAt] element.
+     */
+    override fun subPathOf(createdAt: TimeTicket): String? {
+        return elements.subPathOf(createdAt)
+    }
+
+    // NOTE(7hong13): Original comment from JS SDK is as follows:
+    // `purge` physically purge child element.
```
hackerwins, line 29:
```suggestion
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -1,3 +1,162 @@
...
+    /**
+     * Physically deletes the given [element].
+     */
+    override fun delete(element: CrdtElement) {
+        elements.delete(element)
+    }
+
+    // NOTE(7hong13): Original comment from JS SDK is as follows:
+    // `insertAfter` adds a new node with the value after the given node.
```
hackerwins, line 38:
```suggestion
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -1,3 +1,162 @@
...
+    /**
+     * Returns the element of the given [index].
+     */
+    operator fun get(index: Int): CrdtElement? {
+        val node = elements.getByIndex(index)
+        return node?.value
+    }
+
+    // NOTE(7hong13): Original comment from JS SDK is as follows:
+    // `getPrevCreatedAt` returns the creation time of the previous element of the given element.
```
hackerwins, line 71:
```suggestion
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -1,3 +1,162 @@
...
+    /**
+     * Returns a creation time of the previous node.
+     */
+    fun getPrevCreatedAt(createdAt: TimeTicket): TimeTicket {
+        return elements.getPrevCreatedAt(createdAt)
+    }
+
+    // NOTE(7hong13): Original comment from JS SDK is as follows:
+    // `delete` deletes the element of the given index.
```
hackerwins, line 80:
```suggestion
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/RgaTreeList.kt
```diff
@@ -155,37 +154,45 @@ internal class RgaTreeList {
...
 
-    @VisibleForTesting
-    fun getNodesInListExceptHead(): List<RgaTreeListNode> {
-        return buildList {
-            var curr = dummyHead.next
-            while (curr != null) {
-                add(curr)
-                curr = curr.next
+    // NOTE(7hong13): Original comment from JS SDK is as follows:
+    // `getLastCreatedAt` returns the creation time of last elements.
```
hackerwins, line 182:
```suggestion
```

hackerwins (approved):

We should consider removing elements using `delete` in JavaScript from the user's point of view later. It would be good to discuss it when developing the Document.

```js
const array = [1,2,3];
delete array[0];
```

@skhugh For now, it would be good to merge this PR and revise it considering Kotlin best practices.

yorkie/src/main/kotlin/dev/yorkie/document/time/TimeTicket.kt
skhugh: I checked the code created by Java protobuf compiler and it uses `int` for `uint32`.
so... it seems we are better sticking with `Int` for now?

hackerwins: Thanks for the check. If so, it would be better to use Int for delimiter in Yorkie.

PR #17, opened 2022-10-06
Suggests some improvements on TimeTicket and RhtPQMap

skhugh:
Comparing two nullable TimeTicket instances was ugly so I created extension to handle it gracefully.
But IDE cannot auto-import so we need to do it manually.
(`import dev.yorkie.document.time.TimeTicket.Companion.compareTo`)

changed some functions to operator function for better usages.

Also made constructors private for those with `of`/`create`.
(Although I'm wondering if we really need to use `of`/`create` instead of the default constructors)

I have not looked deeply into `RgaTreeList` yet since it seems to be still being modified on [open PR](https://github.com/yorkie-team/yorkie-android-sdk/pull/12).

Fixes #

### Checklist

- [x] Didn't break anything


hackerwins (approved):
Thanks for your contribution.

PR #19, opened 2022-10-06
Improve test coverage

7hong13:
This PR is to improve test coverage based on [the result from previous PR](https://github.com/yorkie-team/yorkie-android-sdk/pull/12#issuecomment-1269249596).

`CrdtArray` has low test coverage since its overriding functions are not implemented yet.

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
```diff
@@ -10,14 +10,21 @@ internal class CrdtArray(
...
     val last
         get() = elements.last.value
 
     val length
         get() = elements.length
 
+    /**
+     * Returns the creation time of the last element.
+     */
+    fun getLastCreatedAt(): TimeTicket {
```
skhugh, line 24: Although it is only my personal preference, I would use property rather than functions when it always returns the same instance and has no parameters.
(same for `RgaTreeList.getHead`)

skhugh (approved):
I left a couple comments but they are not something that have to be fixed here :)

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtArray.kt
hackerwins: If there are no problems, it would be better to unify the style. The proposed method is looking neater.

hackerwins (approved):
Thanks for your contribution. Please check the comments suggested by @skhugh.

PR #20, opened 2022-10-07
Implements CrdtObject

7hong13:
Implements `CrdtObject` which will be used in `CrdtRoot` and other classes.

It is implemented based on `CRDTObject` in `yorkie-js-sdk`.
I marked `NOTE` on comments which are different from original ones.

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtObject.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtObject.kt
@@ -13,8 +13,10 @@ internal class CrdtObject private constructor(
     val rht
         get() = memberNodes.toList()
 
+    // NOTE(7hong13): the original comment from js-sdk is as follows:
+    // `getKeys` returns array of this object.
     /**
-     * Returns array of this object.
+     * Returns the array of keys in this object.
      */
     val keys
         get() = map { it.first }
--- a/yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtObject.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtObject.kt
@@ -10,8 +10,8 @@ internal class CrdtObject private constructor(
     createdAt: TimeTicket,
     private val memberNodes: RhtPQMap<CrdtElement>,
 ) : CrdtContainer(createdAt) {
-    val Rht
-        get() = memberNodes
+    val rht
+        get() = memberNodes.toList()
 
     /**
      * Returns the sub path of the given element.
@@ -32,7 +32,7 @@ internal class CrdtObject private constructor(
     /**
      * Sets the given element of the given key.
      */
-    fun set(key: String, value: CrdtElement): CrdtElement? {
+    operator fun set(key: String, value: CrdtElement): CrdtElement? {
         return memberNodes.set(key, value)
     }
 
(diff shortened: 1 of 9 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtObject.kt
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * [CrdtObject] represents object datatype, but unlike regular JSON, it has time
+ * tickets which is created by logical clock.
+ */
+internal class CrdtObject private constructor(
+    createdAt: TimeTicket,
+    private val memberNodes: RhtPQMap<CrdtElement>,
+) : CrdtContainer(createdAt) {
+    val Rht
```
7hong13, line 13: maybe it's better to make `memberNodes` public, but not sure

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtObject.kt
```diff
@@ -0,0 +1,128 @@
...
+        get() = memberNodes
```
7hong13, line 14: maybe it's better to make `memberNodes` public, but not sure

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtObject.kt
skhugh, line 13:
```suggestion
    val rht
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtObject.kt
```diff
@@ -0,0 +1,128 @@
...
+     * Physically deletes the given [element].
+     */
+    override fun delete(element: CrdtElement) {
+        memberNodes.delete(element)
+    }
+
+    /**
+     * Sets the given element of the given key.
+     */
+    fun set(key: String, value: CrdtElement): CrdtElement? {
```
skhugh, line 35: let's make this operator fun :)

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtObject.kt
```diff
@@ -0,0 +1,128 @@
...
+     * Removes the element of the given key and execution time.
+     */
+    fun removeByKey(key: String, executedAt: TimeTicket): CrdtElement {
+        return memberNodes.removeByKey(key, executedAt)
+    }
+
+    /**
+     * Returns the value of the given key.
+     */
+    fun get(key: String): CrdtElement {
```
skhugh, line 56: let's make this operator fun :)

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtObject.kt
skhugh: I think it would be better to keep `memberNodes` private but provide a property returning `Iterable<RhtPQMap.RhtPQMapNode<T>>` type. 
this way we can hide mutability.

hackerwins (approved):
Thanks for your contribution and for correcting the wrong comment.

I think @skhugh can check the best practices in Kotlin.

skhugh (approved):

PR #24, opened 2022-10-19
implement JsonArray

skhugh:
implements JsonArray.
also fills out some JsonObject's methods that were left out before.

Fixes #

### Checklist

- [x] Didn't break anything


skhugh:
while working on this I felt there were a couple of things we could consider for improvements.

- rename Primitive to CrdtPrimitive and create JsonPrimitive;
since we don't have union types in Kotlin, using Primitive for both Json and Crdt container could cause chaos.
I think it would be better to each have independent primitive types for clarity.
- make JsonArray and CrdtArray implement MutableIterable or MutableCollection instead immutable one;
well.. they are mutable collections after all :)

PR #29, opened 2022-10-24
Implement toJson, deepCopy, stringEscape, etc.

7hong13:
1. Implemented `toJson`, `deepCopy`, `toSortedJson` functions which were left as `TODO`.
(`toJS` is not implemented yet. It is only for testing, so I guess it is not urgent.)
2. Fixed some minor issues to match with `yorkie-js-sdk`.
3. Added some test codes.
For `stringEscape` function and its test code, I referred to [iOS version of it](https://github.com/yorkie-team/yorkie-ios-sdk/pull/22)

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtElement.kt
```diff
@@ -32,7 +32,7 @@ internal abstract class CrdtElement(
 
     var removedAt: TimeTicket? = removedAt
         set(value) {
-            if (value > createdAt && value > field) {
+            if (value >= createdAt && value > field) {
```
7hong13, line 35: @skhugh this condition has some issues when setting dummyHead in RgaTreeList, so I fixed it.

yorkie/src/main/kotlin/dev/yorkie/util/SplayTreeSet.kt
```diff
@@ -50,7 +50,7 @@ internal class SplayTreeSet<V>(lengthCalculator: LengthCalculator<V>? = null) {
         }
         var node: Node<V> = root
         while (true) {
-            if (target <= node.leftWeight) {
+            if (node.left != null && target <= node.leftWeight) {
```
hackerwins, line 53: 👍

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtElement.kt
skhugh: I actually made custom setter to replace `remove` function.
we should delete one of them.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
```diff
@@ -46,7 +47,7 @@ internal class CrdtRoot(val rootObject: CrdtObject) {
         val subPaths = mutableListOf<String>()
         while (pair.parent != null) {
```
skhugh, line 48: how about chaning like below to remove ugly `!!`s :)
```suggestion
        while (true) {
            val parent = pair.parent ?: break
```

yorkie/src/main/kotlin/dev/yorkie/document/json/EscapeString.kt
```diff
@@ -0,0 +1,22 @@
+package dev.yorkie.document.json
+
+/**
+ * Escapes string.
+ */
+fun escapeString(str: String): String {
+    return str.map {
```
skhugh, line 7: I think `kotlin.text.replace` is the better option here.

skhugh:
I think `toJS` is not needed for Android SDK.
also `toSortedJson` is used for test code only so I prefer removing them (there should be simpler ways to check equality of JSON contents in Kotlin)

7hong13:
@hackerwins as I mentioned in the daily scrum, I removed `toSortedJson`. I will revert it if removing them can cause some issues.

hackerwins (approved):
Thanks for your contribution.

PR #30, opened 2022-10-24
Implement Trie related functions

7hong13:
Added `Trie` which is used in `Document.update` function. 

Basically I used `Trie` from [Apache Commons Collections](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/trie/PatriciaTrie), but added `findPrefixes` function since it's not a supported in Apache.
(I declared `findPrefixes` as a top level function, but not sure whether it is a good approach.)

I also added test codes based on [yorkie-js-sdk version of it](https://github.com/yorkie-team/yorkie-js-sdk/blob/main/test/unit/util/trie_test.ts).

Fixes #

### Checklist

- [x] Didn't break anything


skhugh (approved):
LGTM

PR #31, opened 2022-10-26
implement Client

skhugh:
implements Client.
a lot is left commented because we need converters before implementing them.
also I might change public suspend functions in the future to non-suspend functions so Client can have its own scope to handle concurrency better.

Fixes #

### Checklist

- [x] Didn't break anything


7hong13 (approved):

hackerwins (approved):
Thanks for your contribution.

PR #34, opened 2022-10-27
Converter test codes

7hong13:
Test codes on conversion between Protobuf format and model format.

I set all the `typealias` as `internal`. (I will change it as `private` if it is more proper to use.) 

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/api/CommonConverter.kt
```diff
@@ -5,9 +5,9 @@ import com.google.protobuf.ByteString
 import com.google.protobuf.kotlin.toByteString
 
 internal fun ByteString.toHexString(): String {
-    return BaseEncoding.base16().lowerCase().encode(toByteArray())
+    return BaseEncoding.base16().encode(toByteArray())
 }
 
 internal fun String.toDecodedByteString(): ByteString {
```
7hong13, line 11: I removed `lowercase()` since it seemed like to cause errors when capital letters are given as input.
but not sure it is a right solution.

skhugh (approved):

yorkie/src/main/kotlin/dev/yorkie/api/ChangeConverter.kt
```diff
@@ -69,8 +73,8 @@ internal fun PBChangePack.toChangePack(): ChangePack {
         documentKey = documentKey,
         checkPoint = checkpoint.toCheckPoint(),
         changes = changesList.toChanges(),
-        snapshot = snapshot,
-        minSyncedTicket = minSyncedTicket.toTimeTicket(),
+        snapshot = if (snapshot.isEmpty) null else snapshot,
```
skhugh, line 76: we can do `snapshot.takeUnless { it.isEmpty }`

yorkie/src/main/kotlin/dev/yorkie/api/ChangeConverter.kt
```diff
@@ -69,8 +73,8 @@ internal fun PBChangePack.toChangePack(): ChangePack {
...
+        minSyncedTicket = if (hasMinSyncedTicket()) minSyncedTicket.toTimeTicket() else null,
```
skhugh, line 77: there is kotlin extension `minSyncedTicketOrNull` :)

hackerwins (approved):
Thanks for your contribution.

PR #35, opened 2022-10-27
refactor client and document

skhugh:
use converters in Client.
use single background thread for Client and Document
change suspend functions to non-suspend and return deferred.
separate status from Client.Event and hold them in StateFlow's.

I'm not quite sure if returning `Deferred` is a good idea here. 
I might change them to just use `launch` instead... 🤔
will decide on this when we start working on examples.

test codes are to be added next.

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/api/CommonConverter.kt
```diff
@@ -1,13 +0,0 @@
-package dev.yorkie.api
-
-import com.google.common.io.BaseEncoding
-import com.google.protobuf.ByteString
-import com.google.protobuf.kotlin.toByteString
-
-internal fun ByteString.toHexString(): String {
```
skhugh, line 7: @7hong13 I've created `ActorID` converters in `TimeConverter` instead.

yorkie/src/main/kotlin/dev/yorkie/api/ElementConverter.kt
```diff
@@ -135,13 +139,11 @@ internal fun CrdtElement.toPBJsonObject(): PBJsonElement {
 internal fun CrdtObject.toPBJsonObject(): PBJsonElement {
     val crdtObject = this
     return jSONElement {
-        PBJsonObject.newBuilder().apply {
+        jsonObject = jSONObject {
```
skhugh, line 142: @7hong13 there were a few codes that didn't use kotlin dsl version, so I've changed them.

hackerwins (approved):

PR #39, opened 2022-11-01
Sample kanban board app implementation

7hong13:
Sample Kanban board implementation.
result:

https://user-images.githubusercontent.com/52355963/200484843-d8820d10-508b-454a-9ecc-89f963582bf5.mov
I changed `TimeTicket` as public and used it as a key for `JsonArray`, but it is just temporary try to sync with other SDKs.
we should change the structure later to improve its usability.

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
```diff
@@ -128,12 +128,16 @@ internal class CrdtRoot(val rootObject: CrdtObject) {
...
+        try {
+            removedElementSetByCreatedAt.forEach { createdAt ->
+                val pair = elementPairMapByCreatedAt[createdAt] ?: return@forEach
+                if (pair.element.isRemoved && pair.element.removedAt <= executedAt) {
+                    pair.parent?.delete(pair.element)
+                    count += garbageCollectInternal(pair.element)
+                }
             }
+        } catch (e: Exception) {
+            // TODO(7hong): handle ConcurrentModificationException
```
skhugh, line 140: modifying a collection should not happen inside a loop.
we should use iterator.. all loop a snapshot of a collection.

skhugh: fixed this issue on #43

examples/src/main/kotlin/dev/yorkie/examples/KanbanBoard.kt
```diff
@@ -0,0 +1,196 @@
...
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+data class KanbanColumn(
```
7hong13, line 34: maybe we need new interface to use these (non-crdt) custom classes easily in the document.
since Android SDK doesn't reveal id of the JsonObject to public, I'm not sure how to get particular JsonObject in the document right now... (ex. when adding a new card to the particular column)

examples/src/main/kotlin/dev/yorkie/examples/KanbanBoard.kt
```diff
@@ -0,0 +1,199 @@
...
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import dev.yorkie.document.time.TimeTicket
+
+data class KanbanColumn(
+    val title: String,
+    val cards: List<Card>,
+    var id: TimeTicket = TimeTicket.InitialTimeTicket,
```
skhugh, line 39: using mutable properties should be avoided when we are using Compose.

skhugh (approved):

hackerwins (commented):

PR #41, opened 2022-11-02
use StringBuilder to avoid multiple String concats on toJson.

skhugh:
use StringBuilder to avoid multiple String concats on toJson.

Fixes #

### Checklist

- [x] Didn't break anything


hackerwins (approved):

PR #42, opened 2022-11-02
fix concurrent modification on gc

skhugh:
fix concurrent modification on gc.
I used iterator for `textWithGarbageSetByCreatedAt`, but snapshot for `removedElementSetByCreatedAt` because gc flow is quite complex here.

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
@@ -15,9 +15,9 @@ import dev.yorkie.util.YorkieLogger
  */
 internal class CrdtRoot(val rootObject: CrdtObject) {
     private val elementPairMapByCreatedAt =
-        hashMapOf(rootObject.createdAt to CrdtElementPair(rootObject))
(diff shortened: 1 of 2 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
```diff
@@ -128,21 +128,23 @@ internal class CrdtRoot(val rootObject: CrdtObject) {
...
             val pair = elementPairMapByCreatedAt[createdAt] ?: return@forEach
             if (pair.element.isRemoved && pair.element.removedAt <= executedAt) {
                 pair.parent?.delete(pair.element)
                 count += garbageCollectInternal(pair.element)
             }
         }
 
-        textWithGarbageSetByCreatedAt.forEach { createdAt ->
-            val pair = elementPairMapByCreatedAt[createdAt] ?: return@forEach
+        val textGarbageIterator = textWithGarbageSetByCreatedAt.iterator()
```
7hong13, line 139: 👍🙏

hackerwins (approved):

PR #49, opened 2022-11-18
Implement JsonCounter and change CrdtCounter structure

7hong13:
Implemented JsonCounter and changed CrdtCounter a bit.
(more test code to be added on JsonCounter)

changes on CrdtCounter
- `Double` type is not supported.
-  Int and Long overflow is allowed.
- the counterType is determined right after its creation and cannot be changed.

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/api/ElementConverter.kt
```diff
@@ -250,8 +248,10 @@ internal fun PBJsonElementSimple.toCrdtElement(): CrdtElement {
         )
         PBValueType.VALUE_TYPE_INTEGER_CNT,
         PBValueType.VALUE_TYPE_DOUBLE_CNT,
-        PBValueType.VALUE_TYPE_LONG_CNT,
```
skhugh, line 253: I think you meant to delete the double type here :)

skhugh:
since we are only allowing `Int` and `Long` for `Counter`, `sanitized` seems overcomplicated.
having two `invoke`s for each input type seems to be the better choice now.

skhugh (approved):

hackerwins (approved):
Thanks for your contribution.

PR #56, opened 2022-12-15
Change input format of gRPC target address.

7hong13:
Change input format of gRPC target address.

It is required to provide a custom name resolver for `https` URI scheme. ([ref](https://github.com/grpc/grpc-java/issues/3319#issuecomment-321105118))
For now, it'd be better to use host name and port instead.
I might implement the custom name resolver later.

I will update the [document](https://yorkie.dev/docs/android-sdk#creating-a-client) soon as well.

Fixes #

### Checklist

- [x] Didn't break anything


hackerwins (approved):

PR #63, opened 2023-01-04
Implements Rht

7hong13:
Implements Rht.

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/document/crdt/Rht.kt
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * [Rht] is a replicated hash table by creation time.
+ * For more details about RHT:
+ * @link http://csl.skku.edu/papers/jpdc11.pdf
+ */
+internal class Rht : Iterable<Rht.Node> {
+    private val nodeMapByKey = mutableMapOf<String, Node>()
+
+    val entrySet: List<Entry>
```
7hong13, line 14: I converted the below original function `toObject()` from JS-SDK to `entrySet` in Android-SDK, but I'm not sure if this is the best way to handle it.

```js
  /**
   * `toObject` returns the object of this hashtable.
   */
  public toObject(): Record<string, string> {
    const obj: Record<string, string> = {};
    for (const [key, node] of this.nodeMapByKey) {
      obj[key] = node.getValue();
    }

    return obj;
  }
```

skhugh: I think property name should be `entries` or return `Set`.

skhugh (approved):

PR #67, opened 2023-01-11
few miscellaneous fixes.

skhugh:
this PR includes few fixes that I found needed while trying yorkie on a text editor app.
made clientID publicly accessible as it is needed to check if the remote changes are from peers or not.
marked `root` and `clone` volatile and changed `getRoot` to `suspend` function for thread-safety.
`TextValue.toString` was returning a data class structure which was not an intended behavior.

Fixes #

### Checklist

- [x] Didn't break anything


hackerwins (approved):
Thanks for your contribution.

7hong13 (approved):

PR #71, opened 2023-01-13
fix where deepCopy was not actually deep coping elements.

skhugh:
fix where deepCopy was not actually deep coping elements.

Fixes #

### Checklist

- [ ] Added relevant tests or not required
- [ ] Didn't break anything


hackerwins (approved):
Thanks for your contribution.

PR #74, opened 2023-02-02
Fix presence inconsistency issues

7hong13:
Fixed the presence inconsistency issues mentioned in this [PR](https://github.com/yorkie-team/yorkie-js-sdk/pull/440).

I used Channel to fix this issue, but not sure it is the right way to handle it.

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -361,7 +368,7 @@ public class Client @VisibleForTesting internal constructor(
             document.applyChangePack(pack)
             attachments[document.key] = Attachment(document, !isManualSync)
             runWatchLoop()
-            true
+            presenceMapInitEvent.receive()
```
skhugh, line 371: `Channel` can be tricky to use when there are multiple consumers.
If multiple `attachAsync`'s are requested there would be multiple consumers trying to `receive` from the channel.
And since there is no guarantee of the order of watch document responses, the wrong attach request could receive the init event.
We could use a `flow`, and filter out other attach requests by checking the document keys.
or we could make `runWatchLoop` suspend until first watch document response is received. (we need to check if it is guaranteed that first reponse will always have `initialization` )

Also, there is no need to wait for the watch document response for manual synced documents.

7hong13:
the work will be done at [this PR](https://github.com/yorkie-team/yorkie-android-sdk/pull/75).

PR #78, opened 2023-02-10
few improvements on text type implementation

7hong13:
- override toString() in RgaTreeSplit, CrdtText, JsonText
- implement delete(), empty() in JsonText
- fix ActorID to have lowercase string value

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/json/JsonText.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/json/JsonText.kt
@@ -23,6 +23,9 @@ public class JsonText internal constructor(
     internal val values
         get() = target.values
 
+    public val length: Int
+        get() = target.length
+
     /**
      * Edits this [JsonText] with the given [content] and [attributes].
      */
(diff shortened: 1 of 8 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/json/JsonText.kt
```diff
@@ -105,13 +108,31 @@ public class JsonText internal constructor(
...
+     * Deletes the text in the given range.
+     */
+    public fun delete(fromIndex: Int, toIndex: Int): Boolean {
+        return edit(fromIndex, toIndex, "")
+    }
+
+    /**
+     * Makes the text empty.
+     */
+    public fun empty(): Boolean {
```
skhugh, line 121: I think `clear` would be better to align with Kotlin's API's.

skhugh (approved):

hackerwins (approved):
Thanks for your contribution.

PR #83, opened 2023-02-17
Fix invalid indexOf SplayTree with single node

skhugh:
from https://github.com/yorkie-team/yorkie/pull/470

> Fix invalid index of SplayTree with a single node.
> 
> This PR addresses an issue with the indexOf in SplayTree, which was failing to calculate the correct index when there was only one node in the tree. The conditional statement has been updated to account for this scenario.

Fixes #

### Checklist

- [x] Didn't break anything


hackerwins (approved):
Thanks for your contribution.

PR #86, opened 2023-03-06
Implement the select feature in the sample app.

7hong13:
- Implement the select feature in the sample app so that it can show the selected blocks by peers.

https://user-images.githubusercontent.com/52355963/223011828-cea0540b-eef6-4bd8-9d79-ecb014a1a3ba.mov

- add toDocKey() for integration tests. (https://github.com/yorkie-team/yorkie-js-sdk/pull/465)
related to: https://github.com/yorkie-team/yorkie-ios-sdk/pull/61

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/examples/texteditor/src/main/java/com/example/texteditor/EditorViewModel.kt
+++ b/examples/texteditor/src/main/java/com/example/texteditor/EditorViewModel.kt
@@ -68,7 +68,9 @@ class EditorViewModel(val client: Client) : ViewModel(), YorkieEditText.TextEven
         viewModelScope.launch {
             client.peerStatus.collect { peerStatus ->
                 val peers = peerStatus.map { it.actorId }
-                _removedPeers.emit(peerSelectionInfos.keys.filterNot { actorID -> actorID in peers })
+                _removedPeers.emit(
+                    peerSelectionInfos.keys.filterNot { actorID -> actorID in peers },
+                )
             }
         }
     }
--- a/examples/texteditor/src/main/java/com/example/texteditor/EditorViewModel.kt
+++ b/examples/texteditor/src/main/java/com/example/texteditor/EditorViewModel.kt
@@ -68,9 +68,7 @@ class EditorViewModel(val client: Client) : ViewModel(), YorkieEditText.TextEven
         viewModelScope.launch {
             client.peerStatus.collect { peerStatus ->
                 val peers = peerStatus.map { it.actorId }
-                _removedPeers.emit(
-                    peerSelectionInfos.keys.filterNot { actorID -> actorID in peers },
-                )
+                _removedPeers.emit(peerSelectionInfos.keys.filterNot { it in peers })
             }
         }
     }
--- a/examples/texteditor/src/main/java/com/example/texteditor/EditorViewModel.kt
+++ b/examples/texteditor/src/main/java/com/example/texteditor/EditorViewModel.kt
@@ -5,7 +5,6 @@ import androidx.lifecycle.viewModelScope
 import dev.yorkie.core.Client
 import dev.yorkie.document.Document
 import dev.yorkie.document.crdt.TextChange
-import dev.yorkie.document.crdt.TextChangeType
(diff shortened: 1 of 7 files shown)


examples/texteditor/src/main/java/com/example/texteditor/EditorViewModel.kt
```diff
@@ -23,24 +22,26 @@ class EditorViewModel(private val client: Client) : ViewModel(), YorkieEditText.
...
-                    _textChanges.emit(it)
-                }
+        changes.filterNot { it.actor == clientID }.forEach {
+            viewModelScope.launch {
+                _textChanges.emit(it)
             }
+        }
     }
 
+    var configurationChanged = false
```
skhugh, line 34: is this needed?
it seems it is never set to `true`

examples/texteditor/src/main/java/com/example/texteditor/MainActivity.kt
```diff
@@ -1,14 +1,24 @@
...
 import androidx.lifecycle.lifecycleScope
 import androidx.lifecycle.viewmodel.initializer
 import androidx.lifecycle.viewmodel.viewModelFactory
 import com.example.texteditor.databinding.ActivityMainBinding
 import dev.yorkie.core.Client
+import dev.yorkie.document.crdt.TextChange
+import dev.yorkie.document.crdt.TextChangeType
+import dev.yorkie.document.time.ActorID
 import kotlinx.coroutines.launch
+import java.util.Random
```
skhugh, line 21: there is a kotlin version of `Random`

examples/texteditor/src/main/java/com/example/texteditor/MainActivity.kt
```diff
@@ -21,12 +31,14 @@ class MainActivity : AppCompatActivity() {
...
         }
     }
+    private val peerSelectionInfos = mutableMapOf<ActorID, PeerSelectionInfo>()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         binding = ActivityMainBinding.inflate(layoutInflater)
         setContentView(binding.root)
         binding.textEditor.textEventHandler = viewModel
+        viewModel.configurationChanged = savedInstanceState != null
```
skhugh, line 41: since there is possibility of remote events getting lost during the process of configuration change, it would be better to disable the view's automatic data restoration by Android and initialize the EditText's text with `Document` on `onCreate`.

It is also possible to change shared flow to channel to guarantee all events to be collected, but considering the characteristics of a real-time collaboration, I think using `Document` as a SSOT is the saftest way to handle this situation.

examples/texteditor/src/main/java/com/example/texteditor/MainActivity.kt
```diff
@@ -21,6 +31,7 @@ class MainActivity : AppCompatActivity() {
             }
         }
     }
+    private val peerSelectionInfos = mutableMapOf<ActorID, PeerSelectionInfo>()
```
skhugh, line 34: I think this data would be better if handled by the view model

examples/texteditor/src/main/java/com/example/texteditor/MainActivity.kt
```diff
@@ -61,10 +61,10 @@ class MainActivity : AppCompatActivity() {
             launch {
                 viewModel.client.peerStatus.collect {
                     val peers = it.map { peer -> peer.actorId }
-                    peerSelectionInfos.keys.forEach { actorID ->
+                    viewModel.peerSelectionInfos.keys.forEach { actorID ->
                         if (actorID !in peers) {
                             binding.textEditor.text?.removePrevSpan(actorID)
-                            peerSelectionInfos.remove(actorID)
+                            viewModel.removeDetachedPeerSelectionInfo(actorID)
```
skhugh, line 67: I think this logic should belong to view model

skhugh (approved):

PR #87, opened 2023-03-08
Elaborate on the value of the PeersChanged event

skhugh:
Made improvements on how Client provides changes in peer presences.
It now emits each type of peer change events with a collection of an affected peers.

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/androidTest/kotlin/dev/yorkie/core/ClientTest.kt
```diff
@@ -147,13 +154,17 @@ class ClientTest {
...
+
+            val status = peerStatus.entries.first { it.key == client1.requireClientId() }
+            assertEquals(mapOf("k2" to "v2"), status.value.data)
             assertTrue(
-                peerStatus.first {
-                    it.actorId == client2.requireClientId()
-                }.presenceInfo.data.isEmpty(),
+                peerStatus.entries.first {
+                    it.key == client2.requireClientId()
+                }.value.data.isEmpty(),
```
7hong13, line 167: it still seems to cause failures intermittently, but not sure why🤔

7hong13 (approved):

yorkie/src/androidTest/kotlin/dev/yorkie/core/ClientTest.kt
skhugh: there were two issues :(
Peers class' equality check was only comparing references and not its contents, so `dropWhile` was not working as my intention.
Also, I changed presence property in Attachment to immutable, and always emitted new Attachments when there were any changes, which caused `drop(1)` to be unstable

PR #90, opened 2023-03-16
Implement pause and resume

7hong13:
- Implement the pause and resume functions.
- Pause the real time sync when the Hangul composition starts, and resume when it ends
- Detect the Hangul composition by using Unicode [Hangul Jamo](https://en.wiktionary.org/wiki/Appendix:Unicode/Hangul_Compatibility_Jamo) and [Syllables](https://en.wiktionary.org/wiki/Appendix:Unicode/Hangul_Syllables).

implemented based on the [iOS SDK implementation](https://github.com/yorkie-team/yorkie-ios-sdk/pull/62).

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/core/Attachment.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/core/Attachment.kt
@@ -4,9 +4,9 @@ import dev.yorkie.document.Document
 
 internal data class Attachment(
     val document: Document,
-    var isRealTimeSync: Boolean,
+    val isRealTimeSync: Boolean,
     val peerPresences: Peers = UninitializedPresences,
-    var remoteChangeEventReceived: Boolean = false,
+    val remoteChangeEventReceived: Boolean = false,
 ) {
 
     companion object {
--- a/yorkie/src/main/kotlin/dev/yorkie/core/Attachment.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/core/Attachment.kt
@@ -4,7 +4,7 @@ import dev.yorkie.document.Document
 
 internal data class Attachment(
     val document: Document,
-    val isRealTimeSync: Boolean,
+    var isRealTimeSync: Boolean,
     val peerPresences: Peers = UninitializedPresences,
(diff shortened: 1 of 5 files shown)


yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -469,6 +469,27 @@ public class Client @VisibleForTesting internal constructor(
...
+     */
+    public fun resume(document: Document) {
+        changeRealTimeSetting(document, true)
+    }
+
+    private fun changeRealTimeSetting(document: Document, isRealTimeSync: Boolean) {
+        require(isActive) {
+            "client is not active"
+        }
+        attachments.value[document.key]?.isRealTimeSync = isRealTimeSync
```
7hong13, line 490: if it would be better to throw exceptions for non-existing keys, I'll fix it.

yorkie/src/main/kotlin/dev/yorkie/core/Attachment.kt
```diff
@@ -4,7 +4,7 @@ import dev.yorkie.document.Document
 
 internal data class Attachment(
     val document: Document,
-    val isRealTimeSync: Boolean,
+    var isRealTimeSync: Boolean,
```
skhugh, line 7: having mutable properties for a state could be dangerous, as value changes in such properties do not emit changes to subscirbers, and `isRealTimeSync` is used in a `Flow.map`.
we should keep it immutable and always create a new instance of an Attachment when there are changes.
this is why I changed peerPresences to a immutable map in previous work.

`remoteChangeEventReceived` is a special case, since it is only used inside a while loop, and changes do not need emissions.  but it would be better to make it immutable.

7hong13: thank you for explaining it in detail.
I made `isRealTimeSync` and `remoteChangeEventReceived` immutable: [56aa510](https://github.com/yorkie-team/yorkie-android-sdk/pull/90/commits/56aa51009e6a65328f9a84cf5876b28b53897868)

PR #93, opened 2023-03-21
Add RemoveDocument API

7hong13:
- add RemoveDocument API
- apply WatchDocument API

I have made it throw exceptions when documents have invalid states upon calling functions such as `attachAsync()`, `removeAsync()`, and `detachAsync()`. This approach is based on the implementation of other SDKs.

- https://github.com/yorkie-team/yorkie/issues/464
- https://github.com/yorkie-team/yorkie/pull/484

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
@@ -413,7 +413,7 @@ public class Client @VisibleForTesting internal constructor(
                 return@async true
             }
 
-            document.setStatus(DocumentStatus.Attached)
+            document.status = DocumentStatus.Attached
             attachments.value += document.key to Attachment(
                 document,
                 response.documentId,
@@ -455,7 +455,7 @@ public class Client @VisibleForTesting internal constructor(
             val pack = response.changePack.toChangePack()
             document.applyChangePack(pack)
             if (document.status != DocumentStatus.Removed) {
-                document.setStatus(DocumentStatus.Detached)
+                document.status = DocumentStatus.Detached
             }
             attachments.value -= document.key
             true
@@ -501,7 +501,6 @@ public class Client @VisibleForTesting internal constructor(
                 return@async false
             }
 
-            document.setStatus(DocumentStatus.Removed)
             val request = removeDocumentRequest {
                 clientId = requireClientId().toByteString()
                 changePack = document.createChangePack().toPBChangePack()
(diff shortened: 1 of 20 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -182,6 +201,13 @@ public class Document(public val key: Key) {
         return root.garbageCollect(ticket)
     }
 
+    /**
+     * Updates the status of this [Document].
+     */
+    public fun setStatus(status: DocumentStatus) {
```
skhugh, line 207: I don't think we need a separate set method for status.
let's just make status property's setter public or internal.
(I think it should be internal)

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -461,6 +488,38 @@ public class Client @VisibleForTesting internal constructor(
...
+        return scope.async {
+            require(isActive) {
+                "client is not active"
+            }
+            val attachment = attachments.value[document.key] ?: run {
+                YorkieLogger.e("Client.remove", "document is not attached")
+                return@async false
+            }
+
+            document.setStatus(DocumentStatus.Removed)
```
skhugh, line 504: shouldn't it be updated after we recieve a successful response?

skhugh: it seems updating document's status to removed is also done in `Document.applyChangePack`

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -53,6 +53,9 @@ public class Document(public val key: Key) {
     internal val hasLocalChanges: Boolean
         get() = localChanges.isNotEmpty()
 
+    public var status = DocumentStatus.Detached
```
skhugh, line 56: it should be marked volatile

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -180,9 +182,8 @@ public class Client @VisibleForTesting internal constructor(
     public fun syncAsync(): Deferred<Boolean> {
         return scope.async {
             var isAllSuccess = true
-            attachments.value.map { (_, attachment) ->
-                attachment.document
-            }.asSyncFlow().collect { (document, result) ->
+            val attachments: List<Attachment> = attachments.value.values.toList()
```
skhugh, line 185: we don't need to make values to a list.
we can just change `asSyncFlow`'s receiver's type to `Collection<Attachment> `:)

skhugh (approved):

PR #95, opened 2023-03-23
Implement WatchDocument API

7hong13:
implment WatchDocument API

- https://github.com/yorkie-team/yorkie/pull/484
- https://github.com/yorkie-team/yorkie/pull/491

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
@@ -476,6 +476,7 @@ public class Client @VisibleForTesting internal constructor(
             if (!isActive) {
                 return@async true
             }
+            activationJob.cancel()
             attachments.value.keys.forEach(::cancelWatchJob)
             _streamConnectionStatus.emit(StreamConnectionStatus.Disconnected)
 
--- a/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
@@ -40,7 +40,6 @@ import io.grpc.StatusException
 import io.grpc.android.AndroidChannelBuilder
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
-import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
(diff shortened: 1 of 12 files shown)


yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -221,59 +229,71 @@ public class Client @VisibleForTesting internal constructor(
     }
 
     @OptIn(ExperimentalCoroutinesApi::class)
-    private fun runWatchLoop() {
-        scope.launch(activationJob) {
-            attachments.map { it.filterValues { attachment -> attachment.isRealTimeSync } }
-                .map { it.keys }
+    private fun runWatchLoop(documentKey: Document.Key) {
```
skhugh, line 232: if `runWatchLoop` is invoked on each document's attach, `attachments.map` would be invoked multiple times which is not needed.

maybe we could do something like below so runWatchLoop can be invoked just once on activation.

```
   class WatchJobHolder(val id: String, val job: Job)

    private fun runWatchLoop() {
        scope.launch(activationJob) {
            attachments.map {
                it.filterValues(Attachment::isRealTimeSync)
            }.fold(emptyMap<Document.Key, WatchJobHolder>()) { accumulator, value ->
                (accumulator.keys - value.keys).forEach {
                    accumulator.getValue(it).job.cancel()
                }
                value.entries.associate {
                    val previous = accumulator[it.key]
                    it.key to if (previous?.id == it.value.documentID && previous.job.isActive) {
                        previous
                    } else {
                        previous?.job?.cancel()
                        WatchJobHolder(it.value.documentID, createWatchJob(it.value))
                    }
                }
            }
        }
    }

    private fun createWatchJob(attachment: Attachment): Job {
        return scope.launch {
            service.watchDocument(
                watchDocumentRequest {
                    client = toPBClient()
                    documentId = attachment.documentID
                },
            ).retry {
                it.stackTraceToString()
                _streamConnectionStatus.emit(StreamConnectionStatus.Disconnected)
                delay(options.reconnectStreamDelay.inWholeMilliseconds)
                true
            }.collect {
                _streamConnectionStatus.emit(StreamConnectionStatus.Connected)
                handleWatchDocumentsResponse(attachment.document.key, it)
            }
        }
    }
```

skhugh (approved):

PR #105, opened 2023-06-15
Allow specifying a topic when subscribing to a document

7hong13:
- provide a new API that allows users to subscribe to the document with a specific target path.
- improve the event value of Local/RemoteChange.

you can read the full description for these tasks from the following [link](https://github.com/yorkie-team/yorkie-js-sdk/pull/487#issue-1643050373).

some previously written instrumentation tests are failing occasionally. I'll fix them in a separate PR.

related to https://github.com/yorkie-team/yorkie-js-sdk/pull/487, https://github.com/yorkie-team/yorkie-js-sdk/issues/445

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
@@ -11,8 +11,8 @@ import dev.yorkie.document.change.CheckPoint
 import dev.yorkie.document.crdt.CrdtObject
 import dev.yorkie.document.crdt.CrdtRoot
 import dev.yorkie.document.crdt.ElementRht
+import dev.yorkie.document.json.JsonElement
 import dev.yorkie.document.json.JsonObject
-import dev.yorkie.document.operation.InternalOpInfo
 import dev.yorkie.document.operation.OperationInfo
 import dev.yorkie.document.time.ActorID
 import dev.yorkie.document.time.TimeTicket
@@ -24,7 +24,12 @@ import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.async
 import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.asSharedFlow
+import kotlinx.coroutines.flow.filterNot
+import kotlinx.coroutines.flow.mapNotNull
+import kotlinx.coroutines.flow.shareIn
 import kotlinx.coroutines.withContext
 
 /**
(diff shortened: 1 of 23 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -91,14 +98,82 @@ public class Document(public val key: Key) {
...
+            eventStream.emit(Event.LocalChange(changeInfos))
             true
         }
     }
 
+    /**
+     * Subscribes to events on the document with the specific [targetPath].
+     */
+    public fun events(targetPath: String): SharedFlow<Event> {
+        return targetEventStreams.getOrElse(targetPath) {
```
7hong13, line 114: I’m not sure if this is the common way of dealing shared flow, such as returning shared flow from a function or storing them in a map. 
I’ll refactor it if it seems awkward or if there could be potential issues.

yorkie/src/main/kotlin/dev/yorkie/document/operation/OperationInfo.kt
```diff
@@ -0,0 +1,60 @@
...
+import dev.yorkie.document.crdt.TextWithAttributes
+import dev.yorkie.document.time.TimeTicket
+
+/**
+ * [OperationInfo] represents the information of an operation.
+ * It is used to inform to the user what kind of operation was executed.
+ */
+public sealed class OperationInfo {
+
+    public abstract var path: String
```
7hong13, line 12: I defined `path` as a mutable property because making it read-only seemed to result in clumsy boilerplate code for updating its value.
but I’m open to refactoring it if a better solution exists.

yorkie/src/main/kotlin/dev/yorkie/document/operation/OperationInfo.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    internal var executedAt: TimeTicket = TimeTicket.InitialTimeTicket
```
7hong13, line 14: [js implementation link](https://github.com/yorkie-team/yorkie-js-sdk/blob/9f6e51b5352da2537389fb776c7997af518214cb/src/document/operation/operation.ts#L114)
I changed the property name from `element` to `executedAt`  becauase`element` sounds more appropriate for something like `CrdtElement` or `JsonElement`, while `executedAt` has been used as the property name for `TimeTicket`until now.
(cc. @chacha912 )

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
skhugh: We can use `getOrPut` to skip manually setting the new value at the end, but I don't think we need `targetEventStreams`, as sharing a SharedFlow instance for each path seems unnecessary.
Just returning a filtered Flow without eagerly sharing should be enough.

skhugh (approved):

yorkie/src/main/kotlin/dev/yorkie/document/operation/OperationInfo.kt
chacha912: JS-SDK has been updated to remove `InternalOpInfo`. Could you please review the changes and make the necessary updates accordingly?
https://github.com/yorkie-team/yorkie-js-sdk/pull/566

PR #109, opened 2023-06-19
add actorID to ChangeInfo

7hong13:
https://github.com/yorkie-team/yorkie-js-sdk/pull/508

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
@@ -131,12 +131,12 @@ public class Document(public val key: Key) {
     }
 
     private fun List<Event.ChangeInfo>.filterTargetChangeInfos(targetPath: String) =
-        mapNotNull { (message, operations) ->
+        mapNotNull { (message, operations, actor) ->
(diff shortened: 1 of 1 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -308,6 +308,7 @@ public class Document(public val key: Key) {
         public data class ChangeInfo(
             public val message: String,
             public val operations: List<OperationInfo>,
+            public val actorID: ActorID? = null,
```
skhugh, line 311: do we need this to be nullable?
JAVA version of the `ChangeInfo` protobuf model seems have a non-null `actorID`

7hong13: I just ported the JS SDK, but it seems correct to set it as non-null. 
the android SDK implements `Change` to always have `ChangeID` and its `actorID`, so `ChangeInfo` will have a non-null `actorID`.
[0a0c396](https://github.com/yorkie-team/yorkie-android-sdk/pull/109/commits/0a0c3964503573bda40c0ae4b9ea3db6cadfd24d)

skhugh (approved):

PR #117, opened 2023-07-04
Implements tests for JsonTree

skhugh:
Implemtns tests for JsonTree.
Few fixes were made on CrdtTree.

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
@@ -599,7 +599,9 @@ internal class CrdtTreeNode(
         val node = this
         return clone.apply clone@{
             removedAt = node.removedAt
-            value = node.value
+            if (node.isText) {
+                value = node.value
+            }
             size = node.size
             node._children.map { child ->
                 child.deepcopy().apply {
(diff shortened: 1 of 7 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
```diff
@@ -599,7 +599,9 @@ internal class CrdtTreeNode(
         val node = this
         return clone.apply clone@{
             removedAt = node.removedAt
-            value = node.value
+            if (node.isText) {
```
skhugh, line 602: cloning a non-text node was causing execeptins because of value setter

yorkie/src/main/kotlin/dev/yorkie/document/crdt/TreeInfo.kt
```diff
@@ -11,7 +12,32 @@ public data class TreeNode(
     val children: List<TreeNode>? = null,
     val value: String? = null,
     val attributes: Map<String, String>? = null,
-)
+) {
+
+    override fun toString(): String {
```
skhugh, line 17: I changed `toString` to create a json string.

7hong13 (approved):

PR #118, opened 2023-07-04
Implement tests for Tree Operations event emissions

skhugh:
Implement tests for Tree Operations event emissions.
Also added missing set operator for JsonTree.

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/test/kotlin/dev/yorkie/document/json/JsonTreeTest.kt
+++ b/yorkie/src/test/kotlin/dev/yorkie/document/json/JsonTreeTest.kt
@@ -1,5 +1,7 @@
 package dev.yorkie.document.json
 
+import dev.yorkie.document.Document
+import dev.yorkie.document.Document.Event.LocalChange
 import dev.yorkie.document.change.ChangeContext
 import dev.yorkie.document.change.ChangeID
 import dev.yorkie.document.crdt.CrdtObject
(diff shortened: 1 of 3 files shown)


yorkie/src/test/kotlin/dev/yorkie/document/json/JsonTreeTest.kt
```diff
@@ -300,6 +316,142 @@ class JsonTreeTest {
...
+            assertEquals(
+                "<doc><tc><p><tn>aXb</tn></p></tc></doc>",
+                it.tree().toXml(),
+            )
+
+            it.tree().styleByPath(listOf(0, 0, 0), mapOf("a" to "b"))
+        }.await()
+
+        assertContentEquals(
+            listOf(
```
skhugh, line 433: @7hong13  we need to check whether `from`, `to` indices returned with events are actual values.

yorkie/src/test/kotlin/dev/yorkie/document/json/JsonTreeTest.kt
```diff
@@ -300,6 +316,142 @@ class JsonTreeTest {
...
+                    1,
+                    listOf(0, 0),
+                    listOf(0, 0),
+                    TreeNode("text", value = "X"),
+                    "$.t",
+                ),
+                TreeStyleOpInfo(
+                    0,
+                    0,
+                    listOf(0, -7), // TODO: need to check if these are the actual right values???
```
skhugh, line 369: @7hong13  we need to check whether `fromPath`, `toPath`  returned with events are actual values when editing with index.

7hong13: I'm not sure it can have negative values. I'll look up to it later.

yorkie/src/test/kotlin/dev/yorkie/document/json/JsonTreeTest.kt
7hong13: I'll look it up too.

7hong13 (approved):

yorkie/src/test/kotlin/dev/yorkie/document/json/JsonTreeTest.kt
7hong13: @skhugh 
I think it results in negative numbers since `from` and `to` indexes are invalid.
the range should be on the closing tag, like `it.tree().style(4, 5, mapOf("a" to "b"))` in this case.
then the `OpInfo` should be like below:
```kotlin
TreeStyleOpInfo(
         4,
         5,
         listOf(0),
         listOf(0),
         mapOf("a" to "b"),
         "$.t",
),
```

I examined the JS SDK, and it also yields negative value paths for the same case.

yorkie/src/test/kotlin/dev/yorkie/document/json/JsonTreeTest.kt
```diff
@@ -299,6 +315,142 @@ class JsonTreeTest {
...
+                    4,
+                    4,
+                    listOf(0, 0, 0, 1),
+                    listOf(0, 0, 0, 1),
+                    TreeNode("text", value = "X"),
+                    "$.t",
+                ),
+                TreeStyleOpInfo(
+                    6,
+                    7, // TODO: need to check if these are the actual right values???
```
7hong13, line 443: @skhugh 
`from`, `to` are correct for the given TC since it always indicates the closing tag.

PR #123, opened 2023-07-17
Add tests for GC and IndexTree

7hong13:
- Added more tests for IndexTree.
- Added integration tests for GC.
- Changed protobuf data type of `client_seq` and `delimiter` from `uint32` to `uint64`. (cc. @hackerwins )
    - [Protobuf uses a Java int for the Protobuf type uint32](https://protobuf.dev/programming-guides/proto2/#scalar). It causes issues when the server sends a `delimiter` or `client_seq` larger than `Int.MAX_VALUE`. I replaced all `uint32` with `uint64`, which is converted to a long in Java, so that we can properly express all `uint32` values.
- Removed unnecessary codes from GC in [967de53](https://github.com/yorkie-team/yorkie-android-sdk/pull/123/commits/967de5315de1d4f564bf99c8b208f52064e7bc7c)
    - https://github.com/yorkie-team/yorkie-js-sdk/pull/568

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/proto/src/main/proto/yorkie/v1/resources.proto
+++ b/yorkie/proto/src/main/proto/yorkie/v1/resources.proto
@@ -47,7 +47,7 @@ message Change {
 }
 
 message ChangeID {
-  uint64 client_seq = 1;
+  uint32 client_seq = 1;
   int64 server_seq = 2 [jstype = JS_STRING];
   int64 lamport = 3 [jstype = JS_STRING];
   bytes actor_id = 4;
@@ -292,7 +292,7 @@ message Client {
 
 message Checkpoint {
   int64 server_seq = 1 [jstype = JS_STRING];
-  uint64 client_seq = 2;
+  uint32 client_seq = 2;
 }
 
 message TextNodePos {
@@ -303,7 +303,7 @@ message TextNodePos {
 
 message TimeTicket {
   int64 lamport = 1 [jstype = JS_STRING];
-  uint64 delimiter = 2;
+  uint32 delimiter = 2;
   bytes actor_id = 3;
 }
 
--- a/yorkie/proto/src/main/proto/yorkie/v1/resources.proto
+++ b/yorkie/proto/src/main/proto/yorkie/v1/resources.proto
@@ -47,7 +47,7 @@ message Change {
 }
 
 message ChangeID {
-  uint32 client_seq = 1;
+  uint64 client_seq = 1;
   int64 server_seq = 2 [jstype = JS_STRING];
   int64 lamport = 3 [jstype = JS_STRING];
   bytes actor_id = 4;
@@ -292,7 +292,7 @@ message Client {
 
 message Checkpoint {
   int64 server_seq = 1 [jstype = JS_STRING];
-  uint32 client_seq = 2;
+  uint64 client_seq = 2;
 }
 
 message TextNodePos {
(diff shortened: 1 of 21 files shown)


yorkie/proto/src/main/proto/yorkie/v1/resources.proto
```diff
@@ -47,7 +47,7 @@ message Change {
 }
 
 message ChangeID {
-  uint32 client_seq = 1;
+  uint64 client_seq = 1;
```
7hong13, line 50: As stated in the PR description above, I changed `uint32` to `uint64` since Java does not support `uint32` and automatically converts it to `int32`.  
(If `minSyncedTicket` exceeds `Int.MAX_VALUE`, it is converted incorrectly, and GC does not work properly.)

If changing it can have potential issues, I’ll find another solution for it.
(cc. @hackerwins)

yorkie/src/androidTest/kotlin/dev/yorkie/document/json/JsonTreeTest.kt
```diff
@@ -17,122 +16,116 @@ class JsonTreeTest {
 
     @Test
     fun test_tree_sync_between_replicas() {
-        runBlocking {
```
7hong13, line 20: just removed nested `runblocking`s

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
```diff
@@ -618,7 +617,7 @@ internal data class CrdtTreeNode private constructor(
     @Suppress("FunctionName")
     companion object {
 
-        fun CrdtTreeText(pos: CrdtTreePos, value: String? = null): CrdtTreeNode {
+        fun CrdtTreeText(pos: CrdtTreePos, value: String): CrdtTreeNode {
```
7hong13, line 620: text node has always its text value, so I made it non-nullable.

yorkie/proto/src/main/proto/yorkie/v1/resources.proto
skhugh: This was discussed [here](https://github.com/yorkie-team/yorkie-android-sdk/pull/12#discussion_r988468911).
Since exceeding `Int.MAX_VALUE` is unlikely to happen in real world, I think it's okay to stick with it.
Also I'm not sure having different protobuf models between different platforms is a good idea.
But this is something that should be decided by @hackerwins.

7hong13: It appears that the `minSyncedTicket.delimiter` is sometimes set to the maximum value of uint32 for GC. In the Android SDK, this value is then converted to -1, resulting in nothing being garbage collected.
If this is the only case that exceeding `Int.MAX_VALUE`, we can just consider -1 as `MAX_DELIMITER` without changing protobuf models:
```kotlin
internal fun PBTimeTicket.toTimeTicket(): TimeTicket {
    return TimeTicket(
        lamport = lamport,
        delimiter = if (delimiter == -1) MAX_DELIMITER else delimiter,
        actorID = actorId.toActorID(),
    )
}
```
I think it needs to be checked by @hackerwins.

7hong13: I agree it would be better to use `UInt` for kotlin models without changing protobuf models.
done in [aad775c](https://github.com/yorkie-team/yorkie-android-sdk/pull/123/commits/aad775cd93e1fa6e14d6f25484bfc098f4015446).

skhugh (approved):

PR #124, opened 2023-07-18
Prevent usage of `.` in JsonObject key

7hong13:
- I added a restriction on JsonObject key not to contain the delimiter `.`.
- `getValueByPath()` was not functioning correctly, so I fixed it.

related to https://github.com/yorkie-team/yorkie-js-sdk/pull/569, https://github.com/yorkie-team/yorkie-ios-sdk/pull/86

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,7 @@ jobs:
           disable-animations: true
           script: |
             adb shell pm list packages | grep dev.yorkie.test && adb uninstall dev.yorkie.test || true;
+            adb logcat *:E &
             ./gradlew yorkie:createDebugAndroidTestCoverageReport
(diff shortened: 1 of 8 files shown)


7hong13:
It seems like the Github Action for instrumentation tests is not working well.
I haven't found a solution yet, but I'm investigating the issue.

hackerwins:
@7hong13 Currently, the protocol is changed in the main branch to introduce document/presence batch and that seems to be the cause of the failure.

https://github.com/yorkie-team/yorkie/pull/582

How about fixing the server version to v0.4.4 until starting v0.4.5?

https://github.com/yorkie-team/yorkie-android-sdk/blob/main/docker/docker-compose.yml#L25

PR #126, opened 2023-07-21
Update TreeEdit protobuf model and improve JsonTree functionality

7hong13:
#### What this PR does / why we need it?
- updated TreeEdit protobuf model and related code.
- fixed misbehaviors of `JsonTree`.
- added `pathToIndex()` in `JsonTree`.

#### Any background context you want to provide?
some tests will fail in the CI due to its downgraded server version, but I checked they work fine in the latest server.

#### What are the relevant tickets?
- https://github.com/yorkie-team/yorkie-js-sdk/pull/570
- https://github.com/yorkie-team/yorkie-js-sdk/pull/576
- https://github.com/yorkie-team/yorkie-js-sdk/pull/567

Fixes #

### Checklist

- [x] Added relevant tests or not required
- [x] Didn't break anything

PR #130, opened 2023-08-14
Move Presence from Client to Document and replace selection with presence

7hong13:
- Move presence from the client to the document.
- Replace selection handling with presence.

This PR builds upon the changes introduced in the following PR:
- https://github.com/yorkie-team/yorkie-android-sdk/pull/126

- https://github.com/yorkie-team/yorkie-js-sdk/pull/574
- https://github.com/yorkie-team/yorkie-js-sdk/pull/578
- https://github.com/yorkie-team/yorkie-js-sdk/pull/599

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/core/PresenceProxy.kt
```diff
@@ -0,0 +1,32 @@
+package dev.yorkie.core
+
+import dev.yorkie.document.change.ChangeContext
+import dev.yorkie.document.time.ActorID
+
+/**
+ * [PresenceProxy] represents a proxy for the Presence to be manipulated from the outside.
+ */
+public data class PresenceProxy internal constructor(
```
skhugh, line 9: I think we should remove 'Proxy' from the class name, since the users shouldn't have to know whether it is a proxy or not.

yorkie/src/main/kotlin/dev/yorkie/core/PresenceProxy.kt
```diff
@@ -0,0 +1,32 @@
...
+    private val changeContext: ChangeContext,
+    private val presences: Presences,
+    private val actorID: ActorID,
+) {
+
+    public fun set(data: Map<String, String>) {
```
skhugh, line 15: `set` could be a little misleading as it may imply that previous presence data would be overriden by new data map.
Maybe we should use `put` instead, or we could change the parameter type to collection of pairs of strings.

yorkie/src/main/kotlin/dev/yorkie/core/PresenceProxy.kt
7hong13: I renamed it to `Presence` in [this commit](https://github.com/yorkie-team/yorkie-android-sdk/pull/130/commits/77ab8e1ff2074a08645ed0359f2f98116761f759#diff-5cb2a1100515c19243e7fa5708c7f146e511e17229a2057b7dc2febbc1ab092a).

yorkie/src/main/kotlin/dev/yorkie/core/PresenceProxy.kt
7hong13: I changed the function name as `put` in [this commit](https://github.com/yorkie-team/yorkie-android-sdk/pull/130/commits/77ab8e1ff2074a08645ed0359f2f98116761f759#diff-5cb2a1100515c19243e7fa5708c7f146e511e17229a2057b7dc2febbc1ab092a).

7hong13:
In [77ab8e1](https://github.com/yorkie-team/yorkie-android-sdk/pull/130/commits/77ab8e1ff2074a08645ed0359f2f98116761f759), several adjustments have been made to maintain `presences` as a `StateFlow` and to handle flow emission correctly. If storing `presences` in a map seems sufficient or there is a better way to improve it, I will consider making additional changes.

skhugh (approved):

PR #151, opened 2024-01-19
Ensure PresenceChangedEvent is triggered after `presences` emission

7hong13:
It ensures that the `PresenceChanged` event is posted after the `presences` flow emission.

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
@@ -46,6 +46,7 @@ import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.mapNotNull
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.withContext
+import kotlinx.coroutines.withTimeoutOrNull
 
 /**
  * A CRDT-based data type.
@@ -144,7 +145,7 @@ public class Document(public val key: Key, private val options: Options = Option
             }
             if (change.hasPresenceChange) {
                 val presence = _presences.value[actorID] ?: return@async false
-                eventStream.emit(createPresenceChangedEvent(actorID, presence))
+                publishPresenceEvent(createPresenceChangedEvent(actorID, presence))
             }
             true
         }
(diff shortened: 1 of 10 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -369,21 +370,25 @@ public class Document(public val key: Key, private val options: Options = Option
...
+                is Others.Unwatched -> event.changed.actorID !in presences
+                is Others.PresenceChanged -> {
+                    val actorID = event.changed.actorID
+                    actorID !in presences || event.changed.presence == presences[actorID]
+                }
             }
-
-            else -> {}
+        }
+        withTimeoutOrNull(3_000) {
```
skhugh, line 390: do we need the timeout?

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -369,21 +370,25 @@ public class Document(public val key: Key, private val options: Options = Option
...
+            presences.first(predicate)
```
skhugh, line 391: is it possible for presence events to be changed as below very fast that `publishPresenceEvent` for `publishPresenceEvent` is called after it is already unwatched?

```
Watched -> PresenceChanged -> Unwatched
```

skhugh (commented):
I'm not sure if this would be consistent when presences events occur very rapidly.
Maybe we should add test codes to be sure.

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
7hong13: It is logically unnecessary, but I added it just in case of an unexpected issue which can lead to an indefinite delay in the `first()` function.
I'll remove it if it looks redundant.

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
7hong13: I'm not 100% sure that situation will never happen.
I'll fix it to ensure that the `PresenceChanged` event is ignored when it is already unwatched.

PR #153, opened 2024-01-23
Enable users to choose which dispatchers for Client and Document

skhugh:
Enabled users to choose which dispatcher to use for Client and Document, rather than using a fixed single-threaded dispatcher.

While applying Yorkie to a text editor it was found that using a background thread for document updates could cause problems as users can input text while remote changes for document are being applied.
It is very hard to control Android's text input as it is mostly controlled by Android framework, therefore using a main thread for document updates should be recommended.
But for usecases where developer can control the sequence of data changes in detail, using a background thread can still be advantageous.

I also changed Client and Doucment to Closeable as default single-thread dispatcher should be closed.

Fixes #

### Checklist

- [ ] Added relevant tests or not required
- [ ] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
@@ -29,10 +29,13 @@ import dev.yorkie.document.time.TimeTicket
 import dev.yorkie.document.time.TimeTicket.Companion.InitialTimeTicket
 import dev.yorkie.util.YorkieLogger
 import dev.yorkie.util.createSingleThreadDispatcher
+import java.io.Closeable
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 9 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -50,9 +53,16 @@ import kotlinx.coroutines.withContext
 /**
  * A CRDT-based data type.
  * We can represent the model of the application and edit it even while offline.
+ *
+ * A single-threaded, [Closeable] [dispatcher] is used as default.
+ * Therefore you need to [close] the client, when the client is no longer needed.
```
7hong13, line 58: it's not that crucial, but I think `client` should be updated to `document` in the comment.
```suggestion
 * Therefore you need to [close] the document, when the document is no longer needed.
```

7hong13 (approved):

PR #163, opened 2024-03-21
fix issues with handling presence events

7hong13:
This PR addresses the following issues related to handling presence events:
- corrects the condition for the `Initialized` event publishing
- resolves the blocking of all events due to a specific erroneous event.
- fixes `ConcurrentModificationException` during the presence event queue iteration

Fixes #

### Checklist

- [x] Didn't break anything


yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -105,7 +105,9 @@ public class Document(
     public val presences: StateFlow<Presences> =
         combine(_presences, onlineClients) { presences, onlineClients ->
             presences.filterKeys { it in onlineClients }.asPresences()
-        }.stateIn(scope, SharingStarted.Eagerly, _presences.value.asPresences())
+        }.stateIn(scope, SharingStarted.Eagerly, _presences.value.asPresences()).also {
+            scope.launch { it.collect(::publishPresenceEvent) }
+        }
```
skhugh, line 110:
```
public val presences: StateFlow<Presences> =
    combine(_presences, onlineClients) { presences, onlineClients ->
        presences.filterKeys { it in onlineClients }.asPresences()
    }.onEach {
        publishPresenceEvent(it)
    }.stateIn(scope, SharingStarted.Eagerly, _presences.value.asPresences())
```

maybe `onEach` could be used instead, timing would be a little different but I don't think it would be a problem.

skhugh (approved):

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
7hong13: I initially tried to use `onEach` but encountered failing TCs due to timing issues, so I just retained the current implementation.

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -395,16 +391,17 @@ public class Document(
      * Triggers an event in this [Document].
      */
     private suspend fun publishPresenceEvent(presences: Presences) {
-        val iterator = presenceEventQueue.listIterator()
+        val publishedEvents = mutableListOf<Event.PresenceChange>()
+        val iterator = presenceEventQueue.toList().iterator()
```
skhugh, line 395: since it is changed to copying the list, we could just use forEach for the loop

7hong13: done in [8643e03](https://github.com/yorkie-team/yorkie-android-sdk/pull/163/commits/8643e037748755c89d3190be04cf72bbb02f66c7)

PR #171, opened 2024-04-15
Implement Tree.RemoveStyle

7hong13:
Implement `JsonTree.removeStyle()`.

- https://github.com/yorkie-team/yorkie-js-sdk/pull/745
- https://github.com/yorkie-team/yorkie-js-sdk/pull/755

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeStyleOperation.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeStyleOperation.kt
@@ -14,8 +14,9 @@ internal data class TreeStyleOperation(
     override val parentCreatedAt: TimeTicket,
     val fromPos: CrdtTreePos,
     val toPos: CrdtTreePos,
-    val attributes: Map<String, String>,
     override var executedAt: TimeTicket,
+    val attributes: Map<String, String>? = null,
+    val attributesToRemove: List<String>? = null,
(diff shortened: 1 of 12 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeStyleOperation.kt
```diff
@@ -29,16 +30,33 @@ internal data class TreeStyleOperation(
...
+                        it.from,
+                        it.to,
+                        it.fromPath,
+                        attributesToRemove = it.attributesToRemove.orEmpty(),
+                        path = root.createPath(parentCreatedAt),
+                    )
+                }
+            }
+
+            else -> emptyList()
```
skhugh, line 59: is it assured that only one of `attributes` or `attributesToRemove` will be set?

skhugh (approved):

yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeStyleOperation.kt
7hong13: In the current interface, `style()` can only set `attributes`, and `removeStyle()` can only set `attributesToRemove`. So I think it is logically assured that only one of them will be set.

PR #174, opened 2024-05-10
Improve performance

skhugh:
Improves performance by:
1. Reducing number of class conversions.
2. Reducing number of filtering of tree nodes.
3. Providing a way to set different dispatcher for snapshot operations.

When text grows there can be hundreds of text nodes, and converting & filtering a such large list caused a performance impact.

Fixes #

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/crdt/TreeInfo.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/crdt/TreeInfo.kt
@@ -10,63 +10,48 @@ import dev.yorkie.util.IndexTreeNode.Companion.DEFAULT_TEXT_TYPE
  */
 internal sealed class TreeNode {
     abstract val type: String
-    abstract val childNodes: List<TreeNode>?
-    abstract val value: String?
-    abstract val attributes: Map<String, String>?
 
     fun toJsonTreeNode(): JsonTree.TreeNode {
         return this as JsonTree.TreeNode
-    }
-
-    override fun toString(): String {
-        return if (type == DEFAULT_TEXT_TYPE) {
-            """{"type":"$type","value":"$value"}"""
-        } else {
-            val ssb = StringBuilder(
-                """{"type":"$type","children":[${
-                    childNodes.orEmpty().joinToString(",")
-                }]""",
-            )
-            val attributes = attributes
-            if (attributes?.isNotEmpty() == true) {
-                ssb.append(",")
-                ssb.append(
-                    """"attributes":{${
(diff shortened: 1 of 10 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/crdt/TreeInfo.kt
```diff
@@ -8,23 +8,14 @@ import dev.yorkie.util.IndexTreeNode.Companion.DEFAULT_TEXT_TYPE
...
-    val type: String,
-    val children: List<TreeNode>? = null,
-    val value: String? = null,
-    val attributes: Map<String, String>? = null,
-) {
+internal sealed class TreeNode {
+    abstract val type: String
+    abstract val childNodes: List<TreeNode>?
+    abstract val value: String?
+    abstract val attributes: Map<String, String>?
```
7hong13, line 15: I think it can be refactored to have only `type` as abstract, with the other properties implemented in each subclass.

7hong13 (approved):

yorkie/src/main/kotlin/dev/yorkie/util/IndexTree.kt
```diff
@@ -397,19 +398,31 @@ internal abstract class IndexTreeNode<T : IndexTreeNode<T>>(children: MutableLis
             return parent?.children?.getOrNull(offset + 1)
         }
 
+    private var canUseCachedChildren = false
```
7hong13, line 401: I think it's a good approach to reduce performance overhead, but I'm a bit concerned that updating `canUserCachedChildren` might be overlooked when implementing new methods that update `childrenInternal` later. Adding a comment on this might be helpful, maybe...🤔

yorkie/src/main/kotlin/dev/yorkie/document/crdt/TreeInfo.kt
skhugh: done at [8392c71](https://github.com/yorkie-team/yorkie-android-sdk/pull/174/commits/8392c7143ce7ec0ec881596ee138eb57b0c272ae)

yorkie/src/main/kotlin/dev/yorkie/util/IndexTree.kt
skhugh: Implemented IndexTreeNodeList to handle this better [6c93e13](https://github.com/yorkie-team/yorkie-android-sdk/pull/174/commits/6c93e130f20ea991186932a2487a86b326fbe009)

PR #181, opened 2024-05-16
Reset online clients when stream is disconnected

7hong13:

Fixes #
- https://github.com/yorkie-team/yorkie-js-sdk/pull/796

### Checklist

- [x] Didn't break anything

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
@@ -351,7 +351,7 @@ public class Client @VisibleForTesting internal constructor(
     }
 
     private suspend fun onWatchStreamCanceled(document: Document) {
-        if (document.status == DocumentStatus.Attached) {
+        if (document.status == DocumentStatus.Attached && status.value is Status.Activated) {
             document.publishEvent(
                 Initialized((requireClientId() to document.myPresence).asPresences()),
             )
--- a/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
@@ -291,7 +291,6 @@ public class Client @VisibleForTesting internal constructor(
                     attachment.document.key.documentBasedRequestHeader,
                 ).also {
                     latestStream = it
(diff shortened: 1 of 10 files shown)


yorkie/src/main/kotlin/dev/yorkie/core/Presences.kt
```diff
@@ -18,9 +18,9 @@ public class Presences private constructor(
 
     public operator fun minus(actorID: ActorID): Presences = (map - actorID).asPresences()
 
-    override fun hashCode(): Int = map.hashCode()
-
-    override fun equals(other: Any?): Boolean = map == (other as? Presences)?.map
```
7hong13, line 23: I removed it to ensure that `document.publishPresenceEvent()` is properly invoked whenever a new value is emitted to `document.presences`.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -290,14 +291,14 @@ public class Client @VisibleForTesting internal constructor(
...
                     latestStream = it
+                    attachment.document.publishEvent(StreamConnectionChanged.Connected)
                 }
                 val streamJob = launch(start = CoroutineStart.UNDISPATCHED) {
                     val channel = stream.responseChannel()
                     var retry = 0
                     while (!stream.isReceiveClosed() && !channel.isClosedForReceive) {
                         val receiveResult = channel.receiveCatching()
                         receiveResult.onSuccess {
-                            attachment.document.publishEvent(StreamConnectionChange.Connected)
```
skhugh, line 300: don't we need this line?
if the stream fails then succeed on retry, users would not receive the connected state.

7hong13: I missed it. fixed in [4f46596](https://github.com/yorkie-team/yorkie-android-sdk/pull/181/commits/4f465969d8507f4916feeaeec7e84af1603792aa)

skhugh (approved):

7hong13:
Some tests are failing in the CI, but I checked they're working fine with the real server. I'll investigate these failing tests after merging.

PR #194, opened 2024-06-03
Further reduces instance creations by removing unnecessary toLists

skhugh:
Further reduces instance creations by removing unnecessary toLists

Fixes #

### Checklist

- [x] Didn't break anything

## Summary by CodeRabbit

- **Refactor**
  - Improved internal data handling by changing parameter types from `List` to `Iterable` for better flexibility and performance.
  - Simplified code structure by renaming private variables and updating their usage.
  - Updated function signatures to accept variable arguments instead of lists for better usability.

- **Tests**
  - Enhanced test assertions to accommodate changes in data handling, ensuring robust validation of operations.

### DIFF
--- a/yorkie/src/main/kotlin/dev/yorkie/document/json/JsonTree.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/document/json/JsonTree.kt
@@ -163,19 +163,18 @@ public class JsonTree internal constructor(
     ) {
         if (contents.isNotEmpty()) {
             validateTreeNodes(*contents)
-            if (contents.first().type != DEFAULT_TEXT_TYPE) {
-                contents.forEach { validateTreeNodes(it) }
-            }
         }
 
         val ticket = context.lastTimeTicket
         val crdtNodes = if (contents.firstOrNull()?.type == DEFAULT_TEXT_TYPE) {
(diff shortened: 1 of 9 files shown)


7hong13 (approved):

skhugh:
@7hong13 I also removed `filterIsInstance` to skip creation of new lists, which should be safe because the nodes are validated before the operation.

yorkie/src/main/kotlin/dev/yorkie/document/json/JsonTree.kt
```diff
@@ -152,30 +152,29 @@ public class JsonTree internal constructor(
...
-        contents: List<TreeNode>,
         splitLevel: Int = 0,
+        vararg contents: TreeNode,
     ) {
         if (contents.isNotEmpty()) {
-            validateTreeNodes(contents)
-            if (contents.first().type != DEFAULT_TEXT_TYPE) {
-                contents.forEach { validateTreeNodes(listOf(it)) }
-            }
+            validateTreeNodes(*contents)
```
skhugh, line 165: @7hong13 I made some fixes here.
before it was validating the same nodes already validated just before.
I think it should validate on the children nodes.

7hong13 (approved):

PR #205, opened 2024-06-13
Always recreate stream on failure.

skhugh:
Recreate stream if timed out.

When ClosedReceiveChannelException is received, reusing the same stream caused channel receive to suspend forever.
Therefore we only reused the stream on failure not related to ClosedReceiveChannelException.
But there is suspicion that other failures can also cause it to hang forever.
This PR always creates a new stream on any failures. Also it uses `withTimeout` to prevent such behavior and recreate the stream.

Fixes #

### Checklist

- [x] Didn't break anything

## Summary by CodeRabbit

- **New Features**
  - Enhanced execution control with timeout functionality, improving reliability and performance.
- **Chores**
  - Updated version name to `0.4.24-rc2` for the upcoming release.


7hong13 (approved):

skhugh:
all the test cases are passing on the real server.

PR #228, opened 2025-07-23
Enhance proto buf generation

ntcong1997:
- Issue: Building the app too frequently in a short period causes an error in the :yorkie:bufGenerate task. This happens because the protobuf files are generated every time the app is built, exceeding the generation limit.
Screenshot 2025-07-23 at 15 15 53

- Solution: Only generate protobuf files when there are changes. This avoids unnecessary regeneration and helps stay within the generation limit.

Fixes https://jira.navercorp.com/browse/RTCOLLABPLATFORM-347

### Checklist

- [x] Didn't break anything

## Summary by CodeRabbit

* **Chores**
  * Improved build process for protocol buffer files, enabling faster and more efficient builds.
  * Added new commands to check the status of generated files and to clean up generated code.
  * Enhanced feedback during builds by printing messages when regeneration is triggered.

### DIFF
--- a/yorkie/build.gradle.kts
+++ b/yorkie/build.gradle.kts
@@ -51,8 +51,63 @@ tasks.register<Zip>("stuffZip") {
     from("src/stuff")
 }
 
+/*
+ * === Enhanced Protocol Buffer Generation with Incremental Build Support ===
+ *
+ * This configuration optimizes buf generation to only run when necessary:
+ *
+ * 1. **Incremental Builds**: bufGenerate only runs when proto files or buf config change
+ * 2. **Build Cache**: Generated files are cached and reused when possible
+ * 3. **Rate Limit Protection**: Avoids unnecessary network calls to buf.build
+ *
+ * Available tasks:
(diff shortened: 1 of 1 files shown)


yorkie/build.gradle.kts
```diff
@@ -51,8 +51,63 @@ tasks.register<Zip>("stuffZip") {
...
+            println("Cleaned generated proto files")
+        }
+    }
+}
+
+// Make bufGenerate depend on cleaning when proto files change
+tasks.named("bufGenerate") {
+    // Only clean and regenerate if proto files actually changed
+    inputs.files(fileTree("proto")).skipWhenEmpty()
+    mustRunAfter("cleanGeneratedProto")
```
ngocann, line 110:
```suggestion
    dependsOn("cleanGeneratedProto")
```

ngocann: Should we use dependsOn instead of mustRunAfter? The current approach only orders tasks if both are run, but doesn't guarantee cleaning happens.

- mustRunAfter = "IF both tasks run, use this order"
- dependsOn = "ALWAYS run clean before generate"

PR #358, opened 2026-07-29
Support initialRoot, TreeEdit history convergence, and skipHistory updates

thiennguyen-naver:
> **RTCOLLABPLATFORM-745**: [[Yorkie Android SDK] Support initialRoot in Client.attachDocument](https://jira.navercorp.com/browse/RTCOLLABPLATFORM-745)
> **RTCOLLABPLATFORM-746**: [[Yorkie Android SDK] Support history-exempt local updates (updateAsync skipHistory)](https://jira.navercorp.com/browse/RTCOLLABPLATFORM-746)

## Summary

- Add attach-time `initialRoot` support matching Yorkie iOS and JS behavior.
- Preserve existing server content and exclude initialization from undo history.
- Make history-generated TreeEdit undo/redo changes converge on remote replicas after concurrent edits.
- Add a history-exempt local update: `updateAsync(skipHistory = true)` mutates and syncs as usual but records no undo entry and preserves the redo stack.
- Retain the existing `attachDocument`/`updateAsync` overloads and TreeEdit protobuf schema.

## Changes

- Add the public `InitialRoot` type alias and a new `attachDocument` overload.
- Initialize only missing keys after applying the authoritative server response.
- Support every value type constructible through `JsonObject`.
- Normalize history TreeEdit positions and rebuilt content onto the queued operation before serialization.
- Keep rebuilt wire content detached from later live-tree mutations and suppress reverse history for ineffective edits.
- Add an additive `skipHistory: Boolean = false` parameter to `Document.updateAsync`; when `true`, the change skips `pushUndo` and `clearRedo` while mutation, validation, event emission, and sync stay unchanged.
- Add atomic-failure, lifecycle-ordering, wire-payload, history, tree-sync, split/merge, convergence, and skip-history undo/redo tests.
- Keep the server protocol, dependencies, release version, and root `CHANGELOG.md` unchanged.

Initial values and undo/redo results are ordinary Yorkie changes that synchronize through the existing push-pull flow. The TreeEdit fix reuses the existing operation converter and protobuf fields; no server or wire-schema change is required.

`skipHistory` exists for consumer-plugin bookkeeping writes (run-settlement flush, remote reconciliation) that are not user actions; without it, such a write tops the undo stack and the user's undo reverses bookkeeping instead of their own edit. Neither the JS nor the iOS SDK has this capability today — it is a new, local-only flag mirroring the history semantics of remote changes; nothing serializes.

- RTCOLLABPLATFORM-745
- RTCOLLABPLATFORM-746

## Test plan

- [x] Attach a fresh document with initial values and confirm undo/redo history is empty.
- [x] Attach another client and confirm existing initial content is preserved and synchronized.
- [x] Make a normal edit, then confirm undo preserves the initial root and redo restores the edit.
- [x] Run the exact two-client TreeEdit scenario and confirm `12 456` after undo and `123 456` after redo on both replicas.
- [x] Verify delete, replace, split, merge, vanished-target, and history-isolation behavior.
- [x] Round-trip the effective history TreeEdit range, rebuilt IDs, content, and split level through protobuf conversion.
- [x] Publish the PR branch SNAPSHOT and run zero-plugins-android's unchanged collaborative-edit connected suite.
- [x] Interleave an ordinary edit, a `skipHistory` write, and another ordinary edit; confirm undo reverses only the ordinary edits and the skip write is never visited.
- [x] Confirm a `skipHistory` write preserves a pending redo stack and that a bolded edit un-bolds (not deletes) on undo past a skip write.
- [x] Sync a `skipHistory` mutation to a second attached client and confirm identical convergence to an ordinary update.

### Checklist

- [x] Added relevant JVM and Android instrumented tests.
- [x] Didn't break existing attach, Text history, or ordinary TreeEdit behavior.
- [x] No public API, protobuf schema, server, dependency, release-version, or root-changelog change beyond the approved `initialRoot` API and the additive `updateAsync(skipHistory)` parameter.

### DIFF
--- a/yorkie/src/androidTest/kotlin/dev/yorkie/document/json/JsonTreeUndoTest.kt
+++ b/yorkie/src/androidTest/kotlin/dev/yorkie/document/json/JsonTreeUndoTest.kt
@@ -61,7 +61,7 @@ class JsonTreeUndoTest {
      */
     @Test
     fun test_undo_and_redo_text_delete() {
-        withTwoClientsAndDocuments(syncMode = Manual) { c1, _, d1, _, _ ->
+        withTwoClientsAndDocuments(syncMode = Manual) { c1, c2, d1, d2, _ ->
             d1.updateAsync { root, _ ->
                 root.setNewTree(
                     "tree",
@@ -71,26 +71,36 @@ class JsonTreeUndoTest {
                 )
             }.await()
             c1.syncAsync().await()
+            c2.syncAsync().await()
 
             d1.updateAsync { root, _ ->
                 root.getAs<JsonTree>("tree").edit(2, 4)
             }.await()
             c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(
                 "<root><p>hlo</p></root>",
                 d1.getRoot().getAs<JsonTree>("tree").toXml(),
             )
 
             d1.history.undoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(
                 "<root><p>hello</p></root>",
                 d1.getRoot().getAs<JsonTree>("tree").toXml(),
             )
+            assertEquals(
+                "<root><p>hello</p></root>",
+                d2.getRoot().getAs<JsonTree>("tree").toXml(),
+            )
 
             d1.history.redoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(
                 "<root><p>hlo</p></root>",
-                d1.getRoot().getAs<JsonTree>("tree").toXml(),
+                d2.getRoot().getAs<JsonTree>("tree").toXml(),
             )
         }
     }
@@ -100,7 +110,7 @@ class JsonTreeUndoTest {
      */
     @Test
     fun test_undo_and_redo_text_replace() {
-        withTwoClientsAndDocuments(syncMode = Manual) { c1, _, d1, _, _ ->
+        withTwoClientsAndDocuments(syncMode = Manual) { c1, c2, d1, d2, _ ->
             d1.updateAsync { root, _ ->
                 root.setNewTree(
                     "tree",
@@ -110,6 +120,7 @@ class JsonTreeUndoTest {
                 )
             }.await()
             c1.syncAsync().await()
+            c2.syncAsync().await()
 
             // edit(2, 4, "ELL") deletes [2, 4) = "el" and inserts "ELL",
             // producing "hELLlo".
@@ -117,21 +128,26 @@ class JsonTreeUndoTest {
                 root.getAs<JsonTree>("tree").edit(2, 4, text { "ELL" })
             }.await()
             c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(
                 "<root><p>hELLlo</p></root>",
                 d1.getRoot().getAs<JsonTree>("tree").toXml(),
             )
 
             d1.history.undoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(
                 "<root><p>hello</p></root>",
-                d1.getRoot().getAs<JsonTree>("tree").toXml(),
+                d2.getRoot().getAs<JsonTree>("tree").toXml(),
             )
 
             d1.history.redoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(
                 "<root><p>hELLlo</p></root>",
-                d1.getRoot().getAs<JsonTree>("tree").toXml(),
+                d2.getRoot().getAs<JsonTree>("tree").toXml(),
             )
         }
     }
@@ -708,7 +724,7 @@ class JsonTreeUndoTest {
     }
 
     private fun runSplitUndoRedoCase(splitIdx: Int, afterXml: String) {
-        withTwoClientsAndDocuments(syncMode = Manual) { c1, _, d1, _, _ ->
+        withTwoClientsAndDocuments(syncMode = Manual) { c1, c2, d1, d2, _ ->
             d1.updateAsync { root, _ ->
                 root.setNewTree(
                     "tree",
@@ -718,24 +734,32 @@ class JsonTreeUndoTest {
                 )
             }.await()
             c1.syncAsync().await()
+            c2.syncAsync().await()
             val before = d1.getRoot().getAs<JsonTree>("tree").toXml()
 
             d1.updateAsync { root, _ ->
                 root.getAs<JsonTree>("tree").edit(splitIdx, splitIdx, 1)
             }.await()
             c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(afterXml, d1.getRoot().getAs<JsonTree>("tree").toXml())
 
             d1.history.undoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(before, d1.getRoot().getAs<JsonTree>("tree").toXml())
+            assertEquals(before, d2.getRoot().getAs<JsonTree>("tree").toXml())
 
             d1.history.redoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(afterXml, d1.getRoot().getAs<JsonTree>("tree").toXml())
+            assertEquals(afterXml, d2.getRoot().getAs<JsonTree>("tree").toXml())
         }
     }
 
     private fun runL2SplitUndoRedoCase(splitIdx: Int, afterXml: String) {
-        withTwoClientsAndDocuments(syncMode = Manual) { c1, _, d1, _, _ ->
+        withTwoClientsAndDocuments(syncMode = Manual) { c1, c2, d1, d2, _ ->
             d1.updateAsync { root, _ ->
                 root.setNewTree(
                     "tree",
@@ -745,19 +769,27 @@ class JsonTreeUndoTest {
                 )
             }.await()
             c1.syncAsync().await()
+            c2.syncAsync().await()
             val before = d1.getRoot().getAs<JsonTree>("tree").toXml()
 
             d1.updateAsync { root, _ ->
                 root.getAs<JsonTree>("tree").edit(splitIdx, splitIdx, 2)
             }.await()
             c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(afterXml, d1.getRoot().getAs<JsonTree>("tree").toXml())
 
             d1.history.undoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(before, d1.getRoot().getAs<JsonTree>("tree").toXml())
+            assertEquals(before, d2.getRoot().getAs<JsonTree>("tree").toXml())
 
             d1.history.redoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(afterXml, d1.getRoot().getAs<JsonTree>("tree").toXml())
+            assertEquals(afterXml, d2.getRoot().getAs<JsonTree>("tree").toXml())
         }
     }
 
@@ -769,7 +801,7 @@ class JsonTreeUndoTest {
      */
     @Test
     fun test_undo_cross_boundary_merge_restores_siblings() {
-        withTwoClientsAndDocuments(syncMode = Manual) { c1, _, d1, _, _ ->
+        withTwoClientsAndDocuments(syncMode = Manual) { c1, c2, d1, d2, _ ->
             d1.updateAsync { root, _ ->
                 root.setNewTree(
                     "t",
@@ -780,6 +812,7 @@ class JsonTreeUndoTest {
                 )
             }.await()
             c1.syncAsync().await()
+            c2.syncAsync().await()
             val before = "<doc><p>AB</p><p>CD</p></doc>"
             assertEquals(before, d1.getRoot().getAs<JsonTree>("t").toXml())
 
@@ -787,14 +820,21 @@ class JsonTreeUndoTest {
                 root.getAs<JsonTree>("t").editByPath(listOf(0, 2), listOf(1, 0))
             }.await()
             c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals("<doc><p>ABCD</p></doc>", d1.getRoot().getAs<JsonTree>("t").toXml())
 
             assertTrue(d1.history.canUndo())
             d1.history.undoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(before, d1.getRoot().getAs<JsonTree>("t").toXml())
+            assertEquals(before, d2.getRoot().getAs<JsonTree>("t").toXml())
 
             d1.history.redoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals("<doc><p>ABCD</p></doc>", d1.getRoot().getAs<JsonTree>("t").toXml())
+            assertEquals("<doc><p>ABCD</p></doc>", d2.getRoot().getAs<JsonTree>("t").toXml())
         }
     }
 
--- a/yorkie/src/androidTest/kotlin/dev/yorkie/document/json/JsonTreeUndoTest.kt
+++ b/yorkie/src/androidTest/kotlin/dev/yorkie/document/json/JsonTreeUndoTest.kt
@@ -100,6 +100,10 @@ class JsonTreeUndoTest {
             c2.syncAsync().await()
             assertEquals(
                 "<root><p>hlo</p></root>",
+                d1.getRoot().getAs<JsonTree>("tree").toXml(),
+            )
+            assertEquals(
+                "<root><p>hlo</p></root>",
                 d2.getRoot().getAs<JsonTree>("tree").toXml(),
             )
         }
@@ -139,12 +143,20 @@ class JsonTreeUndoTest {
             c2.syncAsync().await()
             assertEquals(
                 "<root><p>hello</p></root>",
+                d1.getRoot().getAs<JsonTree>("tree").toXml(),
+            )
+            assertEquals(
+                "<root><p>hello</p></root>",
                 d2.getRoot().getAs<JsonTree>("tree").toXml(),
             )
 
             d1.history.redoAsync().await()
             c1.syncAsync().await()
             c2.syncAsync().await()
+            assertEquals(
+                "<root><p>hELLlo</p></root>",
+                d1.getRoot().getAs<JsonTree>("tree").toXml(),
+            )
             assertEquals(
                 "<root><p>hELLlo</p></root>",
                 d2.getRoot().getAs<JsonTree>("tree").toXml(),
--- a/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
@@ -116,9 +116,16 @@ import dev.yorkie.api.v1.ChannelEvent.Type as PbChannelEventType
  * Initializers for missing root keys when attaching a [Document].
  *
  * Each map key is authoritative: its initializer receives that key and is invoked only when the
- * server-provided root does not contain it. Initial values become ordinary Yorkie changes but are
- * removed from undo history before attachment completes. Concurrent first attachments may both
+ * server-provided root does not contain it. The absence guard checks only the mapped key itself —
+ * an initializer is expected to write only that key; writes to other keys are not guarded against
+ * and are not part of the documented contract. Initial values become ordinary Yorkie changes but
+ * are removed from undo history before attachment completes. Concurrent first attachments may both
  * initialize a missing key and converge through normal CRDT conflict resolution.
+ *
+ * The initializer stays a callback (`suspend JsonObject.(key: String) -> Unit`) rather than a
+ * plain value map because a value cannot express [JsonObject.setNewTree] or
+ * [JsonObject.setNewText] construction; Java consumers that need only primitive/object values can
+ * still use this shape by writing directly to the receiver inside the lambda.
  */
 public typealias InitialRoot = Map<String, suspend JsonObject.(key: String) -> Unit>
 
@@ -1053,10 +1060,22 @@ public class Client(
      * from undo history before the attached event. Concurrent first attachments are not globally
      * serialized; their ordinary CRDT changes converge through normal conflict resolution.
      *
+     * The document is marked [ResourceStatus.Attached] and registered with this client BEFORE the
+     * [initialRoot] initializers run, so a throwing initializer leaves the document attached and
+     * detachable rather than rolled back; the returned deferred still resolves to failure. Document
+     * events emitted by the initializer change are always preceded by the `Attached`
+     * [Document.Event.DocumentStatusChanged] event.
+     *
      * @param initialPresence The initial presence of the client.
      * @param syncMode The synchronization mode of the document.
      * @param schema The schema used to validate the document.
-     * @param disableGC Whether this attachment opts out of tombstone garbage collection.
+     * @param disableGC declares that this attachment will not produce or consume
+     * tombstones. The server skips minVV tracking and omits the response
+     * VersionVector for this client. Use only with Counter or primitive
+     * workloads where no client consumes tombstones; misuse on a document
+     * that uses Tree, Text, or Array deletions leads to undefined GC behavior
+     * on this client. Controls only the wire contract and is distinct from
+     * any local-only [Document] GC pass.
      * @param disablePresence Whether this attachment opts out of presence.
      * @param initialRoot Initializers for root keys absent from the authoritative server response.
      */
@@ -1134,30 +1153,14 @@ public class Client(
                 document.setDisablePresence(response.disablePresence)
                 document.applyChangePack(pack)
 
+                // JS SDK v0.7.16 ordering (packages/sdk/src/client/client.ts:665-793 @ 28a5a42e):
+                // Removed early-return (no clearHistory here — JS parity) → applyStatus(Attached)
+                // → attachment registration → runWatchLoop → initialRoot update → single
+                // unconditional clearHistory. On an initializer throw, the document stays
+                // Attached and registered (detachable), matching JS's no-rollback behavior.
                 if (document.getStatus() == ResourceStatus.Removed) {
-                    document.clearHistory()
                     return@async SUCCESS
                 }
-
-                val initialRootResult = try {
-                    document.updateAsync { root, _ ->
-                        initialRoot.forEach { (key, initializer) ->
-                            if (root.getOrNull(key) == null) {
-                                initializer(root, key)
-                            }
-                        }
-                    }.await()
-                } catch (t: Throwable) {
-                    ensureActive()
-                    Result.failure(t)
-                }
-                if (initialRootResult.isFailure) {
-                    return@async initialRootResult
-                }
-
-                // Clear undo/redo stacks so that initialRoot setup operations
-                // are not reachable via undo. Mirrors JS SDK PR #1238.
-                document.clearHistory()
 
                 document.applyStatus(ResourceStatus.Attached)
                 attachments[documentKey] = Attachment(
@@ -1173,6 +1176,26 @@ public class Client(
                 if (syncMode != SyncMode.Manual && syncMode != SyncMode.Polling) {
                     runWatchLoop(documentKey)
                 }
+
+                val initialRootResult = try {
+                    document.updateAsync { root, _ ->
+                        initialRoot.forEach { (key, initializer) ->
+                            if (key !in root.keys) {
+                                initializer(root, key)
+                            }
+                        }
+                    }.await()
+                } catch (t: Throwable) {
+                    ensureActive()
+                    Result.failure(t)
+                }
+                if (initialRootResult.isFailure) {
+                    return@async initialRootResult
+                }
+
+                // Clear undo/redo stacks so that initialRoot setup operations
+                // are not reachable via undo. Mirrors JS SDK PR #1238.
+                document.clearHistory()
             }
             SUCCESS
         }
--- a/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
+++ b/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
@@ -45,6 +45,7 @@ import dev.yorkie.document.Document.Event.PresenceChanged.MyPresence.Initialized
 import dev.yorkie.document.Document.Event.PresenceChanged.Others
 import dev.yorkie.document.Document.Event.StreamConnectionChanged
 import dev.yorkie.document.Document.Event.SyncStatusChanged
+import dev.yorkie.document.json.JsonObject
 import dev.yorkie.document.presence.P
 import dev.yorkie.document.presence.PresenceInfo
 import dev.yorkie.document.presence.Presences.Companion.asPresences
@@ -110,6 +111,16 @@ import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeoutOrNull
 import okhttp3.OkHttpClient
 import dev.yorkie.api.v1.ChannelEvent.Type as PbChannelEventType
+
+/**
+ * Initializers for missing root keys when attaching a [Document].
+ *
+ * Each map key is authoritative: its initializer receives that key and is invoked only when the
+ * server-provided root does not contain it. Initial values become ordinary Yorkie changes but are
+ * removed from undo history before attachment completes. Concurrent first attachments may both
+ * initialize a missing key and converge through normal CRDT conflict resolution.
+ */
+public typealias InitialRoot = Map<String, suspend JsonObject.(key: String) -> Unit>
 
 /**
  * Client that can communicate with the server.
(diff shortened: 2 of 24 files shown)


yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -315,12 +333,14 @@ public class Document(
             changeID = context.getNextId()
 
             // Push reverse ops for undo
-            val reverseHistoryOps = reverseOps.map { HistoryOperation.Op(it) }
-            if (reverseHistoryOps.isNotEmpty()) {
-                internalHistory.pushUndo(reverseHistoryOps)
-            }
-            if (operationInfos.isNotEmpty()) {
-                internalHistory.clearRedo()
+            if (!skipHistory) {
```
ngocann: **A `skipHistory` write that shifts indices leaves the pending undo/redo entries pointing at the wrong range.**

Tree/Text reverse ops carry *integer offsets* (`TreeEditOperation.undoFromOffset/undoToOffset`, same on `EditOperation`), and `History.reconcileTreeEdit`/`reconcileTextEdit` are only ever invoked from `Document.applyChanges`, gated on `change.id.actor != changeID.actor` — i.e. remote changes only. Ordinary local writes get away without reconciliation because LIFO undo always removes them before reaching deeper entries. A `skipHistory` write is never undone, so its index shift is permanent.

On `<root><p></p></root>`:

1. `updateAsync { tree.edit(1, 1, text { "A" }) }` → pushes U1 with `undoFromOffset = 1, undoToOffset = 2`
2. `updateAsync(skipHistory = true) { tree.edit(1, 1, text { "X" }) }` → `<p>XA</p>`, nothing pushed, nothing reconciled
3. `history.undoAsync()` → pops U1, deletes `[1, 2)` → deletes the bookkeeping **"X"** and leaves the user's **"A"**

The preserved redo stack has the same defect. Suggested fix: after a `skipHistory` change, run its `opInfos` through `internalHistory.reconcileTreeEdit`/`reconcileTextEdit` exactly the way `applyChanges` does for remote changes.

Worth noting that every `skipHistory` test in the PR uses index-neutral operations (object-key `set`, `Tree.style`), so this case isn't covered.

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -315,12 +333,14 @@ public class Document(
...
+                val reverseHistoryOps = reverseOps.map { HistoryOperation.Op(it) }
```
ngocann, line 348: **`skipHistory` discards the reverse ops rather than avoiding them, so the flag isn't actually cheap.**

`change.execute(root, ...)` above has already built `reverseOps`. For tree deletions that means `CrdtTree.edit` deep-copied every removed node (`nodesToBeRemoved.map(CrdtTreeNode::deepCopy)`) and `toReverseOperation` converted them to `TreeNode` snapshots — all thrown away here.

The stated use cases (run-settlement flush, remote reconciliation) are exactly the large-delete-shaped writes that pay the most for this. Threading the flag down into `Change.execute` / `OpSource` so reverse-op generation is skipped at the source would make `skipHistory` cost less than an ordinary update instead of the same.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -1089,13 +1134,31 @@ public class Client(
...
+                                initializer(root, key)
+                            }
+                        }
+                    }.await()
+                } catch (t: Throwable) {
+                    ensureActive()
+                    Result.failure(t)
+                }
+                if (initialRootResult.isFailure) {
+                    return@async initialRootResult
```
ngocann, line 1207: **On initialRoot failure the server-side attachment is already live and is now unreachable.**

`service.attachDocument` returned success further up — the server records this client as attached. Returning here skips `applyStatus(Attached)` and the `attachments[documentKey] = ...` registration, so:

- `detachDocument(document)` throws `ErrDocumentNotAttached` (`attachments[documentKey] ?: throw ...`)
- `client.has(key)` is `false`, so `deactivateAsync()` is the only way to release it
- until then the dangling attachment pins the document's minVV and blocks server-side GC

Either detach on the server before returning the failure, or move initialRoot after registration (the JS SDK ordering) and accept a non-atomic attach.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -1089,13 +1134,31 @@ public class Client(
                 document.setDisablePresence(response.disablePresence)
                 document.applyChangePack(pack)
 
+                if (document.getStatus() == ResourceStatus.Removed) {
+                    document.clearHistory()
+                    return@async SUCCESS
+                }
+
+                val initialRootResult = try {
```
ngocann, line 1194: **This emits `LocalChange` before the document is attached, diverging from the JS SDK ordering.**

`applyStatus(ResourceStatus.Attached)` and `attachments[documentKey] = ...` happen ~20 lines below, and the new ClientTest pins this ordering (`assertIs<LocalChange>(lifecycleEvents.first())`). Inside that window a subscriber reacting to the event sees `getStatus() == Detached`, `client.has(key) == false`, `myPresence` empty, and `client.syncAsync(document)` throwing `ErrDocumentNotAttached`.

The JS SDK runs initialRoot *after* `applyStatus(Attached)` / `attachmentMap.set`, so the PR description's "matching Yorkie iOS and JS behavior" doesn't hold for lifecycle ordering. If the pre-attach placement is deliberate (to get the atomic-failure semantics), the overload's KDoc should say that document events can precede the Attached status.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -1089,13 +1134,31 @@ public class Client(
...
+                    document.updateAsync { root, _ ->
+                        initialRoot.forEach { (key, initializer) ->
+                            if (root.getOrNull(key) == null) {
```
ngocann: **Exception-driven absence check on the attach path.**

`JsonObject.getOrNull` calls `get(key)`, which catches `NoSuchElementException` and rethrows a freshly constructed one, which `getOrNull` then catches again. On a first attach every key is absent, so an initialRoot of N keys builds and throws 2N exceptions (each filling a stack trace) while holding `document.mutex`.

`JsonObject.keys` — which this PR's own test already uses (`"edit" in d2.getRoot().keys`) — gives a non-throwing membership test.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -1089,13 +1134,31 @@ public class Client(
                 document.setDisablePresence(response.disablePresence)
                 document.applyChangePack(pack)
 
+                if (document.getStatus() == ResourceStatus.Removed) {
+                    document.clearHistory()
```
ngocann: **`document.clearHistory()` is now duplicated across both exits.**

Moving the `Removed` check above the initialRoot update required copying this call; the other one is ~20 lines below. Nothing between the two points touches the history stacks, so a single `document.clearHistory()` right after `applyChangePack(pack)` gives identical behavior with one call site.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -111,6 +112,16 @@ import kotlinx.coroutines.withTimeoutOrNull
...
 
+/**
+ * Initializers for missing root keys when attaching a [Document].
+ *
+ * Each map key is authoritative: its initializer receives that key and is invoked only when the
+ * server-provided root does not contain it. Initial values become ordinary Yorkie changes but are
+ * removed from undo history before attachment completes. Concurrent first attachments may both
+ * initialize a missing key and converge through normal CRDT conflict resolution.
+ */
+public typealias InitialRoot = Map<String, suspend JsonObject.(key: String) -> Unit>
```
ngocann, line 130: **Passing the key into the initializer lets it write a key other than the one it's guarded by.**

`mapOf("a" to { this["b"] = 1 })` type-checks: the guard below checks `a`, the initializer writes `b`, and `b` gets overwritten on every attach even when the server already has it — the exact thing initialRoot exists to prevent. JS and iOS take a plain value map (`initialRoot: R`), so the key/value binding can't be broken.

If the callback form is needed for `setNewTree`/`setNewText`, dropping the `key` parameter and having `attachDocument` apply the result to the mapped key restores the invariant.

Separately: `suspend JsonObject.(String) -> Unit` isn't constructible from Java, so Java consumers are limited to the no-initialRoot overload. Worth a note in the KDoc.

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -219,11 +220,28 @@ public class Document(
         fun redoAsync(): Deferred<OperationResult>
     }
 
+    /**
+     * Kept for binary compatibility with code compiled against SDK versions
+     * where [updateAsync] lacked the skipHistory parameter.
+     */
+    @Deprecated(level = DeprecationLevel.HIDDEN, message = "Binary compatibility.")
+    public fun updateAsync(
```
ngocann, line 227: **This PR solves "add a parameter without breaking compiled callers" two different ways.**

Here: `@Deprecated(HIDDEN)` bridge + new parameter inline. In `Client.kt:1030-1045`: a second *public* overload that delegates with `initialRoot = emptyMap()`.

They have different costs — the hidden bridge preserves the `$default` synthetic but breaks Kotlin callers passing the updater positionally; the public overload is source-safe but permanently doubles the API surface and its KDoc (and the new `attachDocument` KDoc has already lost the `disableGC` misuse warning the original carries). Picking one convention and documenting it would keep the next parameter addition predictable.

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
```diff
@@ -219,11 +220,28 @@ public class Document(
...
+     *
+     * @param skipHistory Skips recording this change on the undo/redo history
+     * stacks when true. The change still mutates the document, emits events,
+     * and syncs normally; only the undo entry and redo-stack clearing are
+     * skipped, mirroring how remote changes bypass local history. Defaults
+     * to false, which preserves existing history behavior.
      */
     public fun updateAsync(
         message: String? = null,
+        skipHistory: Boolean = false,
```
ngocann: **Source-breaking for Kotlin callers that pass the updater positionally.**

`DeprecationLevel.HIDDEN` removes the 2-parameter overload from Kotlin overload resolution, so `document.updateAsync("msg", myUpdaterRef)` now only resolves against `(String?, Boolean, suspend ...)` and fails to compile on argument 2. Trailing-lambda call sites — all in-repo and example callers — are fine, so this only hits external consumers, but the binary-compat bridge doesn't cover it.

Appending `skipHistory` after `updater`, or giving the history-exempt path its own function name, would avoid it.

yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeEditOperation.kt
```diff
@@ -134,7 +140,9 @@ internal data class TreeEditOperation(
 
         // Reverse ops are only generated for local and undo/redo operations.
         val reverseOps =
-            if (source == OpSource.Local || source == OpSource.UndoRedo) {
+            if (opInfos.isNotEmpty() &&
```
ngocann: **The two normalizations added here weren't applied to `EditOperation`, so JsonText history keeps the same defects.**

`EditOperation.execute:58` computes `actualFrom/actualTo` from `undoFromOffset/undoToOffset`, but `fromPos`/`toPos` stay `val` and are never updated — and `toPBOperation` serializes `operation.fromPos`/`operation.toPos`. Whenever `reconcileOperation` truncates or collapses the range (cases 3/5/6, a concurrent remote edit overlapping the undo range), the local replica edits the reconciled range while peers get the original one — the same divergence this PR fixes for trees.

`EditOperation:92` also still returns `listOf(toReverseOperation(...))` unconditionally, without the `opInfos.isNotEmpty()` guard added here, so a zero-effect text edit still pushes a dead undo entry.

A two-client JsonText test with a remote edit overlapping the end of a pending undo range before `undoAsync` would confirm it.

yorkie/src/test/kotlin/dev/yorkie/document/DocumentSkipHistoryTest.kt
```diff
@@ -0,0 +1,196 @@
...
+
+        // Two undos exactly consume E1 and E3; E2 is never a target.
+        target.history.undoAsync().await()
+        target.history.undoAsync().await()
+        assertFalse(target.history.canUndo())
+    }
+
+    @Test
+    fun `undo twice reverses E3 then E1 while E2 mutation survives`() = runTest {
+        // Distinct keys per write: each key's own reverse op targets that
```
ngocann, line 67: **This comment describes a real failure mode, and the test routes around it instead of covering it.**

The same-key case is reachable in production: `updateAsync { root["k"] = "1" }` then `updateAsync(skipHistory = true) { root["k"] = "bookkeeping" }` tombstones the element the pending undo entry targets. No reconciliation runs — `History.reconcileCreatedAt` is only called from `executeUndoRedo` for ops inside an undo batch, and a `skipHistory` change produces no reverse ops at all — so `undoAsync()` becomes a silent no-op.

The test at line ~105 shows the mirror image: the preserved redo stack replays `k = "1"` and silently clobbers the bookkeeping write. If both are intended semantics they should be asserted as such; if not, they're bugs.

ngocann (commented):
Automated deep review (max effort). 15 findings, ranked most-severe first as inline comments.

**Blocking, in my view:**
1. `Document.kt` — a `skipHistory` write that shifts document indices is never reconciled into the pending undo/redo stacks, so the next undo deletes the bookkeeping content instead of the user's edit. All `skipHistory` tests here use index-neutral ops, so it's uncovered.
2. `Client.kt` — when an initialRoot initializer fails, the server-side attachment is already live but unreachable: `detachDocument` throws, `has(key)` is false, and only `deactivateAsync()` releases it (blocking GC in the meantime).
3. `TreeEditOperation.kt` — the wire-position normalization and the empty-effect guard were applied to trees but not to `EditOperation`, so JsonText history keeps both defects.

**Also worth a look:** `LocalChange` firing before `Attached` (diverges from the JS ordering the description claims), `skipHistory` paying the full reverse-op cost it then discards, `JsonTreeUndoTest` swapping `d1` assertions for `d2` instead of adding them, and the binary-compat test asserting the bridge method rather than the `$default` synthetic real callers link against.

Nice work on the TreeEdit convergence fix itself — mutating the queued op to the reconciled range before serialization is the right shape, and the round-trip tests pin it well.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

yorkie/src/test/kotlin/dev/yorkie/document/DocumentSkipHistoryTest.kt
```diff
@@ -0,0 +1,196 @@
...
+            target.getRoot().getAs<JsonTree>("tree").toXml(),
+        )
+    }
+
+    @Test
+    fun `pre-skipHistory two-parameter updateAsync stays on the JVM binary surface`() {
+        // given: callers compiled before skipHistory existed link against
+        // updateAsync(message, updater) — kept via a hidden bridge overload.
+        val bridges = Document::class.java.methods.filter {
+            it.name == "updateAsync" && it.parameterCount == 2
```
ngocann: **This asserts the wrong symbol for the guarantee in the test name.**

Kotlin callers compiled before `skipHistory` don't call the 2-arg method — they emit `Document.updateAsync$default(doc, null, updater, 1, null)`, signature `(Document, String, Function2, int, Object)`. Filtering on `parameterCount == 2` matches only the bridge itself, so the test still passes in the scenario that actually breaks callers (e.g. the bridge later losing its default parameter values, which drops the `$default` synthetic and gives real callers a `NoSuchMethodError`).

Asserting a `updateAsync$default` with `parameterCount == 5` on `declaredMethods` would cover the real contract.

yorkie/src/test/kotlin/dev/yorkie/document/DocumentSkipHistoryTest.kt
```diff
@@ -0,0 +1,196 @@
...
+import dev.yorkie.document.json.TreeBuilder.text
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.Assert.assertTrue as jAssertTrue
```
ngocann: **Mixing `kotlin.test` and `org.junit.Assert` in one file.**

This alias only exists because `kotlin.test.assertTrue` is imported at line 8; the file then uses `jAssertTrue` (43, 108, 113) and `assertTrue` (141) interchangeably, plus `org.junit.Assert.assertEquals` while every other test in the package uses `kotlin.test.assertEquals`. A reader has to check which library each call came from to know its argument order.

Dropping the JUnit assertion imports and using `kotlin.test` throughout removes the alias and matches the surrounding tests.

yorkie/src/androidTest/kotlin/dev/yorkie/document/json/JsonTreeUndoTest.kt
```diff
@@ -110,28 +120,34 @@ class JsonTreeUndoTest {
...
                 d1.getRoot().getAs<JsonTree>("tree").toXml(),
             )
 
             d1.history.undoAsync().await()
+            c1.syncAsync().await()
+            c2.syncAsync().await()
             assertEquals(
                 "<root><p>hello</p></root>",
-                d1.getRoot().getAs<JsonTree>("tree").toXml(),
+                d2.getRoot().getAs<JsonTree>("tree").toXml(),
```
ngocann, line 150: **These assertions were switched from `d1` to `d2` rather than having `d2` added.**

In this test both the post-undo (here) and post-redo (~line 150) assertions now target `d2` only, so after the initial edit `d1` is never checked again — a regression that leaves the undoing client in the wrong state passes. `test_undo_and_redo_text_delete` similarly lost its post-redo `d1` assertion.

The other cases in this file (`runSplitUndoRedoCase`, `runL2SplitUndoRedoCase`, `test_undo_cross_boundary_merge_restores_siblings`) correctly assert both replicas — these two should match.

yorkie/src/test/kotlin/dev/yorkie/api/ConverterTest.kt
```diff
@@ -316,6 +317,67 @@ class ConverterTest {
...
+
+        val restore = deleteResult.reverseOps.single() as TreeEditOperation
+        restore.executedAt = ticket(6)
+        restore.execute(root, OpSource.UndoRedo, null)
+        val decodedRestore = listOf(restore.toPBOperation()).toOperations().single()
+            as TreeEditOperation
+
+        assertEquals(restore.fromPos, decodedRestore.fromPos)
+        assertEquals(restore.toPos, decodedRestore.toPos)
+        assertEquals(restore.contents?.map { it.id }, decodedRestore.contents?.map { it.id })
```
ngocann, line 362: **Passes vacuously when both sides are null, which is the regression under test.**

`null == null` succeeds here. If the new `contents = editContents?.map(CrdtTreeNode::deepCopy)` assignment in `TreeEditOperation.execute` were reverted, `restore.contents` would be null and this assertion would still pass — so the test wouldn't catch the loss of rebuilt wire content.

Asserting the list is non-empty first, the way `TreeEditOperationReverseTest` does with `restore.contents.orEmpty().single()`, closes the hole.

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
thiennguyen-naver: Fixed in 8e46113a. The remote reconcile block in `applyChanges` is extracted into `reconcileHistoryEdits(result)`, and `updateAsync(skipHistory = true)` now routes its execution result through it — pending undo/redo tree and text entries are index-reconciled exactly as for a remote change. Your repro is pinned in `DocumentSkipHistoryTest` for both tree and text (insert "A", skipHistory insert "X", undo removes "A" and keeps "X"; redo symmetric).

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
thiennguyen-naver: Fixed in 8ab7b914. Added an internal `OpSource.LocalNoHistory` with a shared `producesReverseOps` predicate; all ten reverse-op-producing operations skip generation for it (the six previously-unconditional ones are now gated too), and `CrdtTree.edit` gained a `captureRemovedNodes` opt-out so the removed-node deep-copy snapshot is skipped when no reverse op will be built. A `skipHistory` change now does strictly less work than an ordinary update. Disclosed side effect: the six previously-unconditional producers also stop building (already-discarded) reverse ops under `Remote`.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
thiennguyen-naver: Fixed in 9d27f54e by adopting the JS ordering (verified against yorkie-js-sdk v0.7.16 `client.ts:665-793`): applyStatus(Attached) → attachment registration → watch loop → initialRoot → clearHistory. On initializer failure the attach deferred still fails, but the document stays attached and registered — `has(key)` is true and `detachDocument(document)` succeeds, releasing the server-side attachment (matches JS, which performs no rollback either). KDoc states the non-atomic failure semantics; a unit ClientTest covers failure → Attached → detachable.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
thiennguyen-naver: Fixed by the same reorder in 9d27f54e — initialRoot now runs after `applyStatus(Attached)` and registration, so `LocalChange` follows the `Attached` status event, matching the JS lifecycle ordering. The ClientTest ordering pin was flipped to assert `Attached` precedes `LocalChange`.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
thiennguyen-naver: Fixed in 9d27f54e — the absence check is now `if (key !in root.keys)`; no exception-driven control flow on the attach path.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
thiennguyen-naver: Resolved by the reorder in 9d27f54e: the Removed early-return no longer calls `clearHistory()` (JS does not either), leaving a single unconditional `clearHistory()` after the initialRoot update — the only code on the path that touches history.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
thiennguyen-naver: Kept the callback form, documented the contract instead (9d27f54e). A plain value map cannot express `setNewTree`/`setNewText`/counters on Android, and any callback with a `JsonObject` receiver can technically write other keys regardless of whether we pass the key parameter — so the invariant is documented rather than type-enforced: the `InitialRoot` KDoc now states an initializer must write only its mapped key (the guard checks the map key only), and notes the Java-consumer limitation you raised.

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
thiennguyen-naver: Unified on one convention in 92c916b3 — the `attachDocument` pattern: a public overload whose new parameter is required (no default), plus the prior public signature delegating to it. The `@Deprecated(HIDDEN)` bridge is gone, and the convention is recorded in the project docs. The dropped `disableGC` misuse warning is restored on the initialRoot `attachDocument` overload KDoc in 9d27f54e.

yorkie/src/main/kotlin/dev/yorkie/document/Document.kt
thiennguyen-naver: Fixed in 92c916b3 — the exact pre-skipHistory 2-param `updateAsync(message, updater)` public overload is restored, so positional callers compile again and its 5-parameter `$default` synthetic is back on the binary. The 3-param overload requires `skipHistory` explicitly (no default), which keeps trailing-lambda overload resolution unambiguous.

yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeEditOperation.kt
thiennguyen-naver: Fixed in 8ab7b914 — `EditOperation.fromPos`/`toPos` are now `var` and are assigned the reconciled `actualFrom`/`actualTo` on undo ops before serialization (mirroring `TreeEditOperation`), and reverse-op generation is guarded with `opInfos.isNotEmpty()`. Pinned by an `EditOperationReverseTest` case, a `ConverterTest` round-trip (serialized range == reconciled range), and the two-client instrumented test you suggested — `UndoRedoTest.test_text_undo_and_redo_converge_after_a_concurrent_overlapping_edit` (remote edit overlapping the end of the pending undo range) passes against a live server with both replicas asserted.

yorkie/src/test/kotlin/dev/yorkie/document/DocumentSkipHistoryTest.kt
thiennguyen-naver: Intended semantics — now asserted and documented (6921cf44, KDoc in 92c916b3): a `skipHistory` write is treated exactly like a remote client's write for history purposes. Index shifts are reconciled (the F1 fix), but there is no createdAt retargeting, because remote changes do not get that either. The test now asserts both directions explicitly: undo after the same-key overwrite is a no-op (the bookkeeping value survives), and redo resurrects the original value via LWW.

yorkie/src/test/kotlin/dev/yorkie/document/DocumentSkipHistoryTest.kt
thiennguyen-naver: Fixed in 6921cf44 — the test now asserts exactly one `updateAsync$default` with `parameterCount == 5` on `declaredMethods`, i.e. the synthetic that pre-skipHistory callers actually link against (the restored 2-param overload's `$default`; the 3-param overload's synthetic has 6 parameters).

yorkie/src/test/kotlin/dev/yorkie/document/DocumentSkipHistoryTest.kt
thiennguyen-naver: Fixed in 6921cf44 — `kotlin.test` assertions throughout; the `org.junit.Assert` imports and the `jAssertTrue` alias are gone (`org.junit` remains only for the JUnit4 annotations). Note: this diverges from the repo-wide JUnit-assertion convention deliberately, to match the surrounding package.

yorkie/src/androidTest/kotlin/dev/yorkie/document/json/JsonTreeUndoTest.kt
thiennguyen-naver: Fixed in 6921cf44 — `d1` assertions restored: post-redo in `test_undo_and_redo_text_delete`, and post-undo + post-redo in `test_undo_and_redo_text_replace`, alongside the existing `d2` assertions. Instrumented suite passes 25/25 against a live server.

yorkie/src/test/kotlin/dev/yorkie/api/ConverterTest.kt
thiennguyen-naver: Fixed in 6921cf44 — `assertFalse(restore.contents.isNullOrEmpty())` now precedes the id comparison, so a reverted contents rebuild can no longer pass vacuously.

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
```diff
@@ -1089,13 +1153,15 @@ public class Client(
...
-                // Clear undo/redo stacks so that initialRoot setup operations
-                // are not reachable via undo. Mirrors JS SDK PR #1238.
-                document.clearHistory()
-
+                // JS SDK v0.7.16 ordering (packages/sdk/src/client/client.ts:665-793 @ 28a5a42e):
+                // Removed early-return (no clearHistory here — JS parity) → applyStatus(Attached)
+                // → attachment registration → runWatchLoop → initialRoot update → single
+                // unconditional clearHistory. On an initializer throw, the document stays
+                // Attached and registered (detachable), matching JS's no-rollback behavior.
                 if (document.getStatus() == ResourceStatus.Removed) {
```
ngocann, line 1175: **Regression against `develop`: the removed-document path lost its history clear.**

On the base branch, [`Client.kt:1094-1096`](https://github.com/yorkie-team/yorkie-android-sdk/blob/develop/yorkie/src/main/kotlin/dev/yorkie/core/Client.kt#L1094-L1096) calls `document.clearHistory()` *before* the `ResourceStatus.Removed` check, so a removed document always had its stacks flushed. This early return now skips it, and `executeUndoRedo` has no removed-status guard — only `isUpdating`.

Reachable via `Document` instance reuse:

1. attach a `Document`, make edits (undo stack now non-empty), detach — `detachDocument` doesn't clear history
2. re-attach the same instance; the server reports the document removed
3. `document.history.canUndo()` is `true`, and `undoAsync()` pops and executes: it mutates the local root of a removed document and appends to `localChanges` a change that can never sync

`develop` masked this with the unconditional clear. The JS-parity note above justifies the *ordering*, but not dropping the clear specifically — JS has no `clearHistory` on that path at all, whereas `develop` deliberately did.

One line restores it:

```kotlin
if (document.getStatus() == ResourceStatus.Removed) {
    document.clearHistory()
    return@async SUCCESS
}
```

Low severity given the narrow trigger, and `attachDocument skips initial root when the server returns a removed document` would still pass either way — so worth an explicit decision rather than leaving it implicit. (Guarding `executeUndoRedo` on `status != Removed` would be the deeper fix, but that's pre-existing and out of scope here.)

ngocann (commented):
Follow-up on the fixes in 8ab7b91..71ba1da — re-reviewed all six commits. 14 of my 15 findings are fixed and the 15th (`InitialRoot` key guard) is declined with a documented rationale, which is fine by me.

The `producesReverseOps` sweep across all ten operation types is more than I asked for and a net win: `Add`/`Set`/`Move`/`Increase`/`ArraySet`/`Remove` were previously building reverse ops even for `OpSource.Remote` and discarding them, and moving `prevCreatedAtForArray!!` inside the guard removes a latent NPE on the remote path. The new two-client instrumented test with a remote edit overlapping the end of a pending undo range also promotes my text-parity finding from "plausible mechanism" to actually demonstrated.

One new regression from the fix commits, inline below.

Two intentional behavior changes I'd like to see explicitly confirmed rather than inferred from the test renames, since both are contracts consumers have to code against:

1. **Attach is no longer atomic.** A failing initializer leaves the document `Attached` with a live watch stream while the deferred resolves to failure, so callers must detach on failure. Matches JS and it's in the KDoc — just confirming it's the intended contract, since it inverts what the previous revision's tests asserted.
2. **`skipHistory` same-element overwrite is remote-like.** Undo becomes a silent no-op and redo replays by LWW, clobbering the bookkeeping write. That's a sharp edge for precisely the consumer-plugin bookkeeping use case the flag exists for; worth confirming the plugin side is fine with it.

Neither blocks from my side. CI is green including the 21m instrumented run.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

yorkie/src/main/kotlin/dev/yorkie/core/Client.kt
thiennguyen-naver: Fixed in 457062e0. The single attach-path `clearHistory()` is hoisted above the `ResourceStatus.Removed` early-return — develop parity restored with one call site (equivalent to your suggested in-branch clear), and the initialRoot-window fix from 798e0fbd is unaffected (the initialRoot update itself now runs as `updateAsync(skipHistory = true)`, so nothing needs clearing after it). Your repro is pinned: the removed-doc ClientTest seeds a pre-attach undo entry on a reused Document instance and asserts `canUndo()`/`canRedo()` are both false after the removed-server attach. Agreed the `executeUndoRedo` removed-status guard is the deeper pre-existing fix — recorded in our backlog rather than folded in here. Instrumented suites re-run green against a live server on the final tree (InitialRootTest 2/2, UndoRedoTest 13/13, SkipHistoryTest 3/3 including the new two-client undo/redo convergence probes, JsonTreeUndoTest 25/25).

thiennguyen-naver:
Confirming the two intended contracts from the follow-up review, explicitly:

1. **Attach is not atomic with initialRoot.** Intended and locked at spec time, matching the JS SDK v0.7.16 ordering (`client.ts:665-793`): a failing initializer leaves the document `Attached` with a live watch stream and the deferred resolves to failure; callers detach on failure. As of 798e0fbd/457062e0, history is cleared before the `Attached` status in all cases, and the initializer change is history-exempt via `skipHistory`, so the failure path no longer leaves stale undo entries either. The KDoc on the initialRoot overload states this contract.

2. **`skipHistory` same-element overwrite is remote-like.** Intended: a `skipHistory` write is treated exactly like a remote client's write for history purposes — index reconciliation, but no createdAt retargeting, so undo of an overwritten entry is a no-op and redo replays by LWW. This is documented on `updateAsync` and asserted in both directions in `DocumentSkipHistoryTest`, plus two-client convergence probes for the write → skipHistory → undo/redo → cross-sync sequence (1299a68f). The LWW-clobber sharp edge for the consumer-plugin bookkeeping flow is flagged to the plugin side for an explicit sign-off; if that surfaces a need for retargeting semantics, it would come as a follow-up spec rather than a change to this contract.

New regression from the fix commits is addressed in 457062e0 (thread reply inline).

thiennguyen-naver:
> [Review 5006007812](https://github.com/yorkie-team/yorkie-android-sdk/pull/358#pullrequestreview-5006007812): *"One new regression from the fix commits, inline below. Two intentional behavior changes I'd like to see explicitly confirmed rather than inferred from the test renames..."*

@ngocann Responding to the follow-up review directly:

- **Inline regression** (removed-document path lost its history clear): fixed in 457062e0 — the single attach-path `clearHistory()` is hoisted above the `ResourceStatus.Removed` early-return (develop parity, one call site), with your reuse repro pinned in the removed-doc ClientTest. Thread replied and resolved.
- **Contract 1 (attach not atomic with initialRoot)**: confirmed intended — JS v0.7.16 ordering; callers detach on failure. Stated in the initialRoot overload KDoc.
- **Contract 2 (skipHistory same-element overwrite is remote-like)**: confirmed intended — undo no-op, redo replays by LWW, documented on `updateAsync` and asserted both directions plus two-client convergence probes (1299a68f). The bookkeeping-clobber sharp edge is with the plugin side for explicit sign-off; any retargeting need would come as a follow-up spec, not a change to this contract.

Full detail in the [contracts comment](https://github.com/yorkie-team/yorkie-android-sdk/pull/358#issuecomment-5393080733) above. Branch head 457062e0, all threads resolved, CI green.

ngocann (approved):

PR #360, opened 2026-08-19
[Android] Sync-up Yorkie JS SDK v0.7.14

thiennguyen-naver:
> **RTCOLLABPLATFORM-754**: [[Android] Sync-up Yorkie JS SDK v0.7.14](https://jira.navercorp.com/browse/RTCOLLABPLATFORM-754)

## Summary
- Ports yorkie-js-sdk v0.7.14 into the Android SDK (stacked on the v0.7.13 sync-up, PR #359 — now merged into `develop`; this PR is rebased onto it)
- Adds identity-preserving restore for Tree undo/redo — undo revives the original node IDs instead of minting copies, so concurrent undo/redo converges and GC stays symmetric
- Fixes the tree-merge family: tombstones move with merge (preserving RGA anchors), chained merges flatten to converge anchored concurrent inserts, style ranges no longer cross a concurrent merge anchor
- Fixes purged text runs being recreated in reverse order on undo after GC
- Tree undo offsets are no longer clamped: an out-of-range stale offset fails loudly (as in JS) instead of silently relocating content (review B1)
- Four tree defects shared with yorkie-js-sdk v0.7.14 (review F1–F4) are ported as-is and pinned as characterization tests, not fixed Android-only — see "Known imported defects" below
- Two upstream commits need no Android change (recorded): Primitive int32→Long promotion (unrepresentable in Kotlin's type system), fast-check counter PBT suite (test-only, new dependency)

## Changes
- `IndexTree.kt`: new `moveChild` — relocates a child updating total size always, visible size only for live nodes; `IndexTreeNodeList.onUnremoved` fixes a stale active-children cache when a node is revived
[...]

PR #361, opened 2026-08-19
[Android] Sync-up Yorkie JS SDK v0.7.15

thiennguyen-naver:
> **RTCOLLABPLATFORM-755**: [[Android] Sync-up Yorkie JS SDK v0.7.15](https://jira.navercorp.com/browse/RTCOLLABPLATFORM-755)

## Summary
- Ports yorkie-js-sdk v0.7.15 (two commits) into the Android SDK. Originally stacked on the v0.7.14 sync PR (#360); rebased onto `develop` after #360 merged (`21b4c1ac`), so the diff is now the v0.7.15 delta only.
- Makes Tree restore/retombstone split-aware for concurrent overlapping undo after GC (JS `7b2ab7a4`, yorkie-js-sdk#1315) — closes the known v0.7.14-parity divergence where a piece straddling a restore-span boundary was skipped, losing content and diverging text-node segmentation across replicas.
- Records the Int-counter wrap fix (JS `bb3c73dc`, yorkie-js-sdk#1312) as a no-code determination: Kotlin `Int` arithmetic already wraps mod-2^32, congruent with the fixed JS/Go for every input (`(a mod 2³²) + (b mod 2³²) ≡ (a+b) mod 2³²`) — pinned by ported tests only.
- No proto/converter/wire change in this release; decode-side validation is unchanged from #360 (a zero-length text span stays accepted for JS/iOS parity — see review follow-ups).

## Changes
- `CrdtTree.kt`: new private `isolateTextRange(piece, from, to)` — splits a piece so a node exactly covering `[from, to)` exists (tree analogue of `RgaTreeSplit.isolateRange`); composes the existing `CrdtTreeNode.split`, so a removed split buffers a pending GC pair (zero diff) and a live split reports metadata overhead. Precondition `pieceStart <= from < to <= pieceEnd` enforced; it and `recreateFromSpan`'s missing-value case throw `YorkieException(ErrInvalidArgument)` (iOS twin's error; `ElementRht`/`RgaTreeList` convention), not an unchecked exception.
- `CrdtTree.restore`: isolate replaces the straddle `break`; returns `TreeRestoreResult(untombstoned, recreated, pendingGcPairs, diff)`, draining born-removed split remainders via `drainPendingGcPairs()` in a `finally` (every path, like the server companion's `defer`). KDoc documents the two upstream `it.skip` known limitations.
- `CrdtTree.retombstone`: isolate replaces the straddle `continue` (live pieces only); returns `(pairs, diff)`; upstream-inherited `maxOf(span.length, 1)` clamp documented in-code.
- `TreeEditOperation.executeRestore`: threads the new outputs by name; registers restore's pending pairs **before** unregistering untombstoned targets (split-born targets walk gc→live correctly, mirroring the Text path); accumulates retombstone/restore diffs + recreated sizes into one `root.acc`.
- Stale "Tree restore never splits or isolates" KDoc rewritten at both sites.
- `CrdtCounter.kt`: unchanged (determination above); wrap invariant pinned in `CrdtCounterTest` (0 + 2^31 → `Int.MIN_VALUE`, -1 + 2^31 → `Int.MAX_VALUE`), `CounterWrapConvergenceTest` (two in-memory replicas, 1 + 5_000_000_000L → 705032705) and instrumented `JsonCounterWrapTest` (same case through server 0.7.15).
- Tests (port of `history_tree_concurrent_test.ts`): `TreeRestoreConcurrentTest` — 12-case overlap matrix (6 relations × {undo, undo+redo}, one shared `Relation` val per relation) with a total GC purge asserted before undo; the same matrix undo+redo **over the wire** (every relayed op through `toPBOperation()`/`toOperations()`) with node-identity convergence; interleaved and reverse-interleaved order variants (segmentation identical to the batched run for all six relations; failures collected per relation via `ErrorCollector`, cancellation propagates); post-undo zero-garbage/docSize exactness; plus 2 `@Ignore` known-limitation twins ported skipped from upstream (runtime-verified genuinely red when enabled). Isolate/ordering/GC-exactness cases in `TreeRestoreConvergenceTest` (pending-pair identity, exact live-split diff, live-straddler isolate inside `restore`, drain-on-throw with typed `YorkieException`); op-level pins in `TreeEditOperationReverseTest` (register-before-unregister order, live-split `docSize.live` accounting); `TreeRestoreConverterTest` parity pin for the zero-length text span; 2 instrumented two-client cases in `JsonTreeRestoreTest` asserting XML **and** node-identity sequence (the after-GC case asserts the purge).
- `docker/docker-compose.yml`, `docker/docker-compose-ci.yml`: server image pin 0.7.14 → 0.7.15 (server companion keeps snapshots byte-parallel).
- **Unrelated to v0.7.15, fixed here because it blocked this PR's CI** — `DocPresenceTest.test_receive_document_events_according_to_presence_changes` (`001874af`): the test attached d1 then d2 without awaiting d1's watch stream and relied on `delay(1000)` before d2's presence put. A late Watch RPC delivers the peer via the stream's `Initialization` frame (`setOnlineClients` + `Initialized`, never `Others.Watched`), and an early put folds into a late `Watched` with no `PresenceChanged`, so the test starved (15s timeout ×3 on CI run 34832086581). Now attaches c1 inside `awaitWatchConnected(d1)` before c2 attaches and waits for the `Watched` event instead of the delay — the shape `test_presence_changed_event_with_multiple_changes` already uses. No SDK code changed.

## Review follow-ups (review 5036530353)
- Blocker (stale #360 base): resolved by the rebase onto `develop`; the #360 fixes (`coerceIn` removal, `traverseAllPreorder`, `isNullOrEmpty` encoder guards, `crossSync(overWire)`) are present on HEAD.
- I1 drain on throw path · I2 live-straddler isolate tested at tree and op level (mutation-verified) · I3 order test made genuinely order-sensitive · I4/I5 total purge asserted (JVM and instrumented) · I6 counter wrap pinned over the wire.
- Minors: `@Ignore` limitations in `restore` KDoc; per-relation failure collection; `isolateTextRange` precondition; named `TreeRestoreResult` access; exact pending-pair identity and live-split diff assertions.

## Review follow-ups (re-review 5205146566)
- R1: the overlap matrix now also runs undo+redo over the wire for all six relations; `JsonTreeRestoreTest` asserts the node-identity sequence, not just XML.
- R2: the `DocPresenceTest` race fix is listed above and in the test plan.
- R3: `isolateTextRange` precondition and `recreateFromSpan` missing value throw `YorkieException(ErrInvalidArgument)`.
- R4: the Android-only `isText && length == 0` decode rejection is **reverted** (JS/iOS accept it; a peer can already delete that character legitimately, so the guard protected nothing and added an Android-only failure to apply a remote change). Pinned as a parity test; upstream draft proposes dropping the shared `max(length, 1)` clamp in all SDKs together.
- Nits: `CancellationException` propagates out of the per-relation collector; one `Relation` val per relation feeds both the list and the explicit tests.

## Test plan
- [x] `./gradlew yorkie:testDebugUnitTest` — 759 tests, 0 failures, 3 skipped (2 upstream it.skip twins + 1 pre-existing); `./gradlew lintKotlin` clean.
- [x] Local Yorkie server 0.7.15 (`docker compose -f docker/docker-compose.yml up -d && ./scripts/config-yorkie-local-server.sh`), then `./gradlew yorkie:connectedDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=dev.yorkie.document.json.JsonTreeRestoreTest,dev.yorkie.document.json.JsonCounterWrapTest,dev.yorkie.document.presence.DocPresenceTest` — 17/17 pass on emulator-5554 (Pixel_10).
- [ ] In a two-client sandbox app, concurrently delete overlapping text ranges in a Tree, sync until GC purges, undo on both — both replicas show the pre-delete content; redo on both returns to the post-delete content.

> Items run automatically by CI (lint, unit tests, coverage) are excluded.


longhoang-lewy (changes requested):
## Review summary

The v0.7.15 port itself is **good** — I verified `isolateTextRange`, the split/GC accounting, the GC-pair ordering and the counter determination independently against JS `7b2ab7a45`, the Go server companion `0f229b6a` (yorkie#1924), and `bb3c73dc4`, and found no convergence or accounting defect in the new code. Requesting changes for one structural blocker plus six test/hardening items.

---

## Blocker — this branch is stacked on a stale, since-rebased copy of #360 and would revert its review fixes

`#360`'s branch was rebased after review; the commit SHAs no longer match (`9e00496` vs this branch's `e695c45` for the same "close v0.7.14 sync-up" commit). #361 branched from the **old** history, so it is missing both `83dc57d` (the `gcPairMap` hashCode fix) and `6df2d96` (all four #360 review-blocker fixes). Verified against the file contents rather than the commit graph:

| #360 fix | base tip (`6df2d96`) | this branch |
|---|---|---|
| B1 — stale-offset clamp removed | 0 × `coerceIn` | **1 × `coerceIn`** (still present) |
| B2 — `traverseAllPreorder` | 2 refs | **0** |
| B4 — `isNullOrEmpty` encoder guards | 4 refs | **0** |

Diffing this head against the current base tip shows net *deletions* in exactly those three files (`OperationConverter.kt` −38, plus `CrdtTree.kt` and `TreeEditOperation.kt`), so merging as-is reintroduces the B1 silent content-misplacement bug and drops the B2/B4 fixes.

A concrete symptom already visible in this PR: `crossSync` here has no `overWire` flag, because that was added in `6df2d96` — so the new tests could not have used the over-the-wire harness even where it would have helped.

**Please rebase onto the current `feat/RTCOLLABPLATFORM-754-sync-js-v0714` tip before this is a merge candidate.** Everything below is against the current head; line numbers will shift after the rebase, but the findings are all in v0.7.15 code that survives it.

Two notes on reading the diff, for whoever reviews next: the 47-file / +6490 GitHub diff is mostly stale-base noise — the real v0.7.15 delta is **955 lines across 10 files** (`git diff e695c45..HEAD`), which is what I reviewed. And the PR body's "no proto/converter/wire change in this release" is **correct**; `resources.proto` shows up in the file list only because of the stale base.

---

## Important

**I1 — `restore` does not drain `pendingGcPairs` on the throw path.** `CrdtTree.kt:1431` drains only on the normal return. The Go companion deliberately uses `defer` for exactly this ("Restore drains buffered pending GC pairs via defer on every path, so a mid-restore `isolateTextRange` error can't leak them into the next operation's drain" — yorkie#1924). *Scenario:* a span whose first piece is a removed straddler buffers a pending pair, then the same span's gap path throws at `requireNotNull(span.value)` (`:1567`) or `spanValue.substring` (`:1571`). The born-removed split node stays tombstoned with its pair stranded in the tree-level buffer — a permanent GC leak if no later `edit`/`style` runs on that tree, or misattributed size accounting if one does. *Fix:* `try { ... } finally { drainPendingGcPairs() }` around the span loop.

**I2 — the live straddler split inside `restore` has no test, and it is the one path that can corrupt `docSize`.** `restore` calls `isolateTextRange` on **every** overlapping piece — `target.isRemoved` is only checked *after* the split (`CrdtTree.kt:1408-1412`). A live piece makes `CrdtTreeNode.split` return non-zero `diff` flowing into `restoreDiff` → `root.acc`. Every new test asserts `DataSize(0, 0)` for restore's diff, i.e. only the removed-split path. Drop either `addDataSizes` call and all tests still pass, yet the split-off live node's meta never enters `docSize.live`; when it is later removed, `registerGCPair` subtracts it, driving `docSize.live` negative and mis-reporting size to the server's quota check. Both replicas drift identically, so convergence assertions cannot catch it.

**I3 — the "undo order" test cannot fail for the reason it exists.** Details inline at `TreeRestoreConcurrentTest.kt:287`.

**I4 — `assertTrue(purged > 0)` does not establish that the run was purged.** Details inline at `TreeRestoreConcurrentTest.kt:87`.

**I5 — the instrumented test never verifies the GC purge that defines it.** Details inline at `JsonTreeRestoreTest.kt:108`.

**I6 — the counter fix is never exercised over the wire.** Upstream put its two-client case in `test/integration/counter_test.ts`, against a real server, precisely because the point is agreeing with the **Go** server's int32 arithmetic. The Kotlin port uses `crossSync`, which passes operations in memory with no protobuf round-trip, and there is no `androidTest` counter case with an out-of-int32 value. Nothing pins that `5_000_000_000L` survives `IncreaseOperation` → protobuf → server → snapshot as a Long rather than being truncated on the wire.

---

## Minor

- **`OperationConverter` accepts `isText && length == 0`**, which `retombstone`'s `maxOf(span.length, 1)` then widens into a one-character re-removal of content the sender never deleted — a peer-controlled content-corruption path. Note this survives the rebase: even with #360's `length < 0` fix, a zero-length text span with `value == ""` passes (`0 == 0`). Suggest `(isText && (length <= 0 || value.length != length))`. Not a divergence today since all three SDKs behave alike, but this is the client that can close it.
- **Document the two `@Ignore`d known limitations in `CrdtTree.restore`'s KDoc**, not only in test comments — that is where the next porter will look. (The `@Ignore`s themselves are legitimate: both map one-to-one to upstream `it.skip` blocks, the reason strings are substantive, and `assertConverged` asserts XML **plus** `identitySequence`, so they would genuinely be red if enabled rather than weakened into green.)
- **Two matrix tests loop all 6 relations inside one `@Test`** (`TreeRestoreConcurrentTest.kt:281`, `:314`), so the first failing relation hides the rest. `@RunWith(Parameterized::class)`, or collect and assert at the end.
- Remaining inline: an unenforced precondition on `isolateTextRange`, positional destructuring of `TreeRestoreResult`, and two test assertions weaker than their messages claim.

---

## Verified clean — no need to re-check

- **`isolateTextRange` arithmetic matches JS exactly.** Both splits are provably in `(0, len)`, so `CrdtTreeNode.split` can never return `null` and the `requireNotNull(right)` at `:1461` cannot fire — `IndexTree.kt:767` returns `null` only for `offset == 0 || offset == visibleSize`, and for a text node `visibleSize == value.length` regardless of tombstone state. Preconditions hold inductively in `restore` (via the `pieceIndex++` guard) and by construction in `retombstone` (`findPiecesOverlapping` guarantees `from < to`). Exact-boundary pieces are not split; no zero-length node is producible.
- **Split/GC accounting is symmetric.** Every split-born node is covered exactly once: removed splits return `DataSize(0,0)` and buffer a pending pair; live splits return real overhead and are `acc`ed. The `gcOnlySize`-register / full-size-unregister asymmetry in `CrdtRoot` is byte-identical to JS `root.ts` at v0.7.15 — inherited parity, not a port bug.
- **GC-pair register-before-unregister ordering is correct** and matches JS `tree_edit_operation.ts`. Traced the split-born-and-untombstoned node: register (new key, `gcOnlySize` → gc) then unregister (gc → live) walks correctly; reversing it would make the unregister a silent no-op and strand a *live* node in `gcPairMap`, i.e. eligible for purge. The `registerGCPair` toggle is not reachable here — `retombstone` adds a pair only when `remove()` returns `true`, and pending pairs are always fresh `cloneText` nodes.
- **`maxOf(span.length, 1)` is genuinely upstream** (`Math.max` in JS, `max` in Go), applied identically in all three implementations, and zero-length spans are not locally producible — so it cannot cause divergence. (The decode-side hole above is a separate, peer-controlled path.)
- **The split path converges.** Boundaries derive only from `span.id.offset`/`span.length` — replicated operation data — never from local piece layout. Replicas with different surviving segmentation reach *finer* but mutually-agreed boundaries, which is why both JS and Kotlin assert on XML rather than sorted JSON.
- **The docker pin bump is required, not cargo-culted.** Server v0.7.15 carries the byte-parallel companion (`0f229b6a`); running the integration suite against 0.7.14 would use the old straddle-`break`/`continue` semantics and produce snapshot mismatches.
- **The counter no-code determination is correct**, and on stronger grounds than the PR body gives. `CrdtCounter.increase` is `value.toInt() + primitiveValue.toInt()` for `Int` and the `Long` analogue for `Long`; since `(a mod 2³²) + (b mod 2³²) ≡ (a+b) mod 2³²`, it is congruent to JS's fixed `bigintToInt32(BigInt(value) + delta)` for **every** input, not just the three sampled. `705032705` is the literal value asserted in upstream `counter_test.ts`, not a Kotlin artefact. Kotlin never had the JS bug because it never had a truncate-the-delta-only branch.
- **Upstream parity is complete.** All 14 tests added in JS `v0.7.14..v0.7.15` (12 matrix + 2 skipped + 1 unit counter + 1 integration counter) have Kotlin counterparts, and I found no Kotlin test pinning an Android-only semantic.

---

## Process

CodeRabbit skipped this PR too, for the same reason as #360 (auto-review disabled on non-default base branches), and the only check is the CLA. That is now two consecutive unreviewed, un-CI'd CRDT changes in the same stack — worth enabling checks on the stacked branches before this chain merges.

<sub>Automated multi-agent review (Claude Code): two independent reviewers over the CRDT/operation layer and the tests/upstream parity, scoped to the true v0.7.15 delta (`e695c45..HEAD`) and verified against JS v0.7.15, the Go server companion, and the PR's own code rather than its description. Findings deduplicated and spot-checked.</sub>

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
```diff
@@ -1090,6 +1320,327 @@ internal data class CrdtTree(
...
+                }
+            }
+        }
+
+        // Splitting a removed straddler buffers born-removed remainders as
+        // pending GC pairs (see CrdtTreeNode.split). The caller registers
+        // these BEFORE unregistering the untombstoned targets, so a target
+        // that was itself a split-born piece is walked gc->live correctly
+        // (mirrors the Text path).
+        val pairs = drainPendingGcPairs()
```
longhoang-lewy, line 1431: **I1 (important) — pending GC pairs leak on the throw path; the Go companion uses `defer` here for exactly this reason.**

This drain runs only on the normal return. Server v0.7.15 (`0f229b6a`, yorkie#1924) deliberately defers it: *"Restore drains buffered pending GC pairs via defer on every path, so a mid-restore `isolateTextRange` error can't leak them into the next operation's drain."*

**Failure scenario.** A span whose first piece is a removed straddler buffers a pending pair via `isolateTextRange` (`:1410`). The same span's gap path then throws — `requireNotNull(span.value)` at `:1567`, or `spanValue.substring` at `:1571`. The born-removed split node stays in the tree, tombstoned, with its GC pair stranded in the tree-level buffer. If no later `edit`/`style`/`removeStyle` runs on that tree the tombstone is never registered and never purged (permanent GC leak); if one does, the pair is registered against the wrong operation's size accounting.

**Fix** — wrap the span loop so the drain runs on every path:

```kotlin
val pairs = try {
    for (span in spans) { /* existing body */ }
    drainPendingGcPairs()
} catch (e: Throwable) {
    drainPendingGcPairs()   // never leak into the next op's drain
    throw e
}
```

(or hoist the loop into a private helper and use a plain `try/finally`). Matching the server's `defer` also keeps the three implementations' failure semantics aligned.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
```diff
@@ -1090,6 +1320,327 @@ internal data class CrdtTree(
...
+            val pieces = findPiecesOverlapping(span.id.createdAt, start, end)
+
+            var cursor = start
+            var pieceIndex = 0
+            while (cursor < end) {
+                val piece = pieces.getOrNull(pieceIndex)
+                val pieceStart = piece?.id?.offset ?: Int.MAX_VALUE
+                val pieceEnd = if (piece != null) pieceStart + piece.value.length else Int.MAX_VALUE
+
+                if (piece != null && pieceStart <= cursor) {
```
longhoang-lewy, line 1408: **I2 (important) — the live-straddler branch here is untested, and it is the only path that can drive `docSize.live` negative.**

`isolateTextRange` is called on **every** overlapping piece; `target.isRemoved` is not checked until `:1412`. When the piece is live, `CrdtTreeNode.split` returns a **non-zero** `diff` that flows through `:1411` into `restoreDiff` and on to `root.acc`.

Every new test asserts `DataSize(0, 0)` for restore's diff (`TreeRestoreConvergenceTest.kt:259`, `:274`, `:299`) — i.e. only the removed-split path. `TreeRestoreConvergenceTest.kt:326` covers a live isolate in **`retombstone`**, not in `restore`.

**Failure scenario.** Drop either `addDataSizes` (here at `:1411`, or the `restoreDiff` accumulation in `TreeEditOperation.executeRestore`) and every unit test still passes — but the split-off live node's metadata never enters `docSize.live`. When that node is later removed, `CrdtRoot.registerGCPair` subtracts its size from `live`, driving `docSize.live` negative and mis-reporting document size to the server's quota check. Both replicas drift *identically*, so no convergence assertion can catch it.

**Fix** — sibling to the `retombstone isolates a live straddling piece` test, but on the restore side:

```kotlin
// single LIVE "0123456789" piece, no prior delete
val result = tree.restore(listOf(TreeRestoreSpan(
    id = CrdtTreeNodeID(insertedAt, 3), isText = true,
    length = 4, value = "3456", nodeType = "text",
)))
assertTrue(result.untombstoned.isEmpty())
assertTrue(result.pendingGcPairs.isEmpty())
assertNotEquals(DataSize(0, 0), result.diff)      // the point of the test
assertEquals("<root>0123456789</root>", tree.toXml())  // visibly unchanged, now 3 pieces
```

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
```diff
@@ -1090,6 +1320,327 @@ internal data class CrdtTree(
...
+     * converge on the same text-node segmentation across replicas (the tree
+     * analogue of [RgaTreeSplit]'s `isolateRange`). A live split's overhead
+     * is a normal live-bucket cost the caller accumulates into its own
+     * `diff`; a removed split buffers a pending GC pair internally via
+     * [CrdtTreeNode.split] (contributing zero here) — the caller must still
+     * drain and register those pairs.
+     *
+     * Requires `pieceStart <= from < to <= pieceEnd`.
+     */
+    private fun isolateTextRange(
```
longhoang-lewy, line 1451: **Minor — the documented precondition is unenforced.**

The KDoc at `:1449` states `pieceStart <= from < to <= pieceEnd`, and both current callers honour it (I verified: `restore` maintains it inductively via the `pieceIndex++` guard, `retombstone` gets it from `findPiecesOverlapping`). But nothing enforces it.

If a future caller passes `from < piece.id.offset`, the left split is silently skipped and the returned node covers content **outside** the span — which `restore` would then un-tombstone, reviving content the undo never deleted. That is a silent convergence bug from a one-line mistake.

```kotlin
require(from in piece.id.offset until to && to <= piece.id.offset + piece.value.length) {
    "isolateTextRange range [$from, $to) escapes piece ${piece.id}"
}
```

Cheap, and turns a silent divergence into a loud failure.

yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeEditOperation.kt
```diff
@@ -165,6 +210,92 @@ internal data class TreeEditOperation(
...
+
+        // 2. Revive (restore) by identity. Isolating a range out of a
+        // straddling piece can split off born-removed remainders as pending
+        // GC pairs; register them FIRST so a split-born untombstoned target
+        // is walked gc->live correctly by the unregister below.
+        val (untombstoned, recreated, restorePairs, restoreDiff) = tree.restore(toRestore)
+        restorePairs.forEach(root::registerGCPair)
+        untombstoned.forEach { node -> root.unregisterGCPair(GCPair(tree, node)) }
+        diff = addDataSizes(diff, restoreDiff)
+        recreated.forEach { node -> diff = addDataSizes(diff, node.dataSize) }
```
longhoang-lewy, line 262: **Minor — positional destructuring of a 4-field data class, two of which are same-shaped lists.**

`val (untombstoned, recreated, restorePairs, restoreDiff) = tree.restore(toRestore)` (just above) binds by position. `TreeRestoreResult` (`CrdtTree.kt:82-87`) has `untombstoned: List<CrdtTreeNode>` and `recreated: List<CrdtTreeNode>` adjacent — reorder those two and this silently swaps with **no type error**, sending the wrong list into the size accumulation on this line.

Prefer named access for the two list fields (`result.untombstoned`, `result.recreated`), or keep the destructuring and add a `// order must match TreeRestoreResult` note next to the declaration. Given `recreated` feeds `root.acc` here and `untombstoned` drives the unregister loop, a silent swap would corrupt both GC state and `docSize` at once.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConcurrentTest.kt
```diff
@@ -0,0 +1,426 @@
...
+    // that. Each relation converges under the reverse order too, and both
+    // orders land on the same final node segmentation.
+    @Test
+    fun `converges on undo of overlapping deletes regardless of undo order`() = runTest {
+        for ((label, r1, r2) in overlapRelations) {
+            val (d1, d2) = seed()
+            val initial = d1.getRoot().getAs<JsonTree>("t").toXml()
+            deleteOverlapping(d1, d2, r1, r2)
+
+            d2.history.undoAsync().await()
```
longhoang-lewy, line 287: **I3 (important) — this test cannot fail for the reason it exists.**

`d1` and `d2` are separate `Document` objects with no shared state, and both `undoAsync` calls complete **before** the single `crossSync` at `:289`. So swapping these two lines changes nothing observable — the undos never interact. The comparison at `:301-305` then asserts that run A equals run B of the same deterministic computation.

Concretely: the test is green whether the order is `d2,d1` or `d1,d2`, so it pins convergence and initial-content restoration per relation (fine, and worth keeping) but the *order-independence* claim in its name and in the `:277-279` comment is vacuous.

**Failure scenario.** Any genuine order-sensitivity in `executeRestore` — e.g. pending-pair registration interacting with a peer's already-restored segmentation — is claimed covered and is not. A regression there ships green.

**Fix** — make it the mirror of the interleaved test at `:314`, which *is* genuinely order-sensitive because the second replica's restore runs against the first's already-restored live segmentation:

```kotlin
d2.history.undoAsync().await()
crossSync(d1, d2)
d1.history.undoAsync().await()
crossSync(d1, d2)
```

then compare that result's `identitySequence` against the forward-interleaved one from `:314`. That is the invariant actually worth pinning.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConcurrentTest.kt
```diff
@@ -0,0 +1,426 @@
...
+        r2: Pair<Int, Int>,
+    ) {
+        d1.updateAsync { root, _ -> root.getAs<JsonTree>("t").edit(r1.first, r1.second) }.await()
+        d2.updateAsync { root, _ -> root.getAs<JsonTree>("t").edit(r2.first, r2.second) }.await()
+        crossSync(d1, d2)
+
+        val vector = maxVectorOf(listOf(actor1, actor2))
+        val purged1 = d1.garbageCollect(vector)
+        val purged2 = d2.garbageCollect(vector)
+        assertTrue(purged1 > 0, "d1 must purge the deleted run before undo")
```
longhoang-lewy, line 87: **I4 (important) — `purged > 0` does not establish that the deleted run was purged.**

This passes if *any single* tombstone is collected. For the overlapping relations both replicas hold several tombstoned pieces after `crossSync`, so if a future GC-eligibility change left one piece behind, `restore` would take the cheap `unremove()` path for it — and the whole 12-case matrix would silently stop exercising the recreate+isolate path this PR exists for, while still passing.

`maxVectorOf(listOf(actor1, actor2))` makes every tombstone eligible, so the exact expectation is reachable:

```kotlin
assertEquals(0, d1.garbageLength, "every tombstone must be purged before undo")
assertEquals(0, d2.garbageLength, "every tombstone must be purged before undo")
assertEquals(DataSize(0, 0), d1.getDocSize().gc)
assertEquals(DataSize(0, 0), d2.getDocSize().gc)
```

If some node is legitimately retained for one of the six relations, assert that exact count rather than `> 0` — the number is the specification.

yorkie/src/androidTest/kotlin/dev/yorkie/document/json/JsonTreeRestoreTest.kt
```diff
@@ -0,0 +1,139 @@
...
+
+            // Concurrent overlapping deletes: d1 deletes "45", d2 deletes the
+            // superset "234567".
+            d1.updateAsync { root, _ -> root.getAs<JsonTree>("tree").edit(4, 6) }.await()
+            d2.updateAsync { root, _ -> root.getAs<JsonTree>("tree").edit(2, 8) }.await()
+
+            // Settle several extra rounds so the server's min-synced version
+            // vector advances past both tombstones and the client-side GC
+            // inside applyChangePack purges them before either undo.
+            repeat(5) {
```
longhoang-lewy, line 108: **I5 (important) — nothing verifies the purge that this test is defined by.**

The whole reason this case exists separately from `test_overlapping_tree_deletes_both_undo_converge` above is that these extra rounds advance the server's min-synced version vector far enough for the client-side GC in `applyChangePack` to purge **before** the undos — as the `:105-107` comment says. No assertion checks that.

If the server's GC config, this `repeat` count, or the min-synced-VV relay changes, this test silently degrades into an exact duplicate of the one above and stays green — while the after-GC recreate path it exists to cover over a real server goes untested.

**Fix** — immediately before the two `undoAsync()` calls at `:124`:

```kotlin
assertEquals(0, d1.garbageLength)
assertEquals(0, d2.garbageLength)
```

`Document.garbageLength` is public (`Document.kt:154`) and already used from `androidTest` in `core/GCTest.kt:51`. Otherwise this test is well built — no `delay`/`Thread.sleep`, teardown handled by `withTwoClientsAndDocuments`.

longhoang-lewy (commented):
Line-anchored companion to my review above. All anchors are in the true v0.7.15 delta (`e695c45..HEAD`), so they survive the rebase onto the current #360 tip even though the line numbers will shift.

Nothing here duplicates #360's findings — the stale-base issue is in the summary review, not inline.

yorkie/src/test/kotlin/dev/yorkie/document/CounterWrapConvergenceTest.kt
```diff
@@ -0,0 +1,46 @@
...
+    @Test
+    fun `converges on the wrapped value when an out-of-int32 Long is applied concurrently`() =
+        runTest {
+            val d1 = Document("test-doc")
+            val d2 = Document("test-doc")
+            d1.setActor(actor1)
+            d2.setActor(actor2)
+
+            d1.updateAsync { root, _ -> root.setNewCounter("age", 1) }.await()
+            crossSync(d1, d2)
```
longhoang-lewy, line 31: **I6 (important) — the counter fix is never exercised over the wire, and the wire is the whole point.**

Upstream put its two-client case in `test/integration/counter_test.ts`, against a real server, precisely because the fix is about agreeing with the **Go** server's int32 arithmetic. `crossSync` passes operations in memory with no protobuf round-trip, and there is no `androidTest` counter case using an out-of-int32 value.

So nothing pins that a `5_000_000_000L` delta survives `IncreaseOperation` → protobuf → server → snapshot as a Long rather than being truncated on the wire. A wire-level truncation would make Android disagree with JS and the server while this test stays green.

**Fix** — keep this as the fast JVM pin, and add an instrumented counterpart (new `JsonCounterWrapTest`, or a case in `core/DocumentTest.kt`) with `withTwoClientsAndDocuments(syncMode = Manual)`: `setNewCounter("age", 1)`, sync both, `increase(5_000_000_000L)` on c1, sync both, then assert `705032705` on **both** documents plus `assertEquals(d1.toJson(), d2.toJson())`.

For the record, the no-code determination itself checks out, and more strongly than the PR body argues: since `(a mod 2³²) + (b mod 2³²) ≡ (a+b) mod 2³²`, Kotlin's `toInt()`-then-add is congruent to JS's fixed `bigintToInt32` for *every* input, not just the three sampled — and `705032705` is the literal value upstream asserts.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConvergenceTest.kt
```diff
@@ -0,0 +1,398 @@
...
+    // removed-split contract).
+    @Test
+    fun `isolateTextRange splits only the left boundary and buffers the remainder`() = runTest {
+        val (tree, span) = buildDeletedRun("0123456789", 3, 9)
+        val narrow = span.narrowedTo(5, 4) // [5,9) of the same tombstoned run
+
+        val result = tree.restore(listOf(narrow))
+
+        assertEquals(listOf("5678"), result.untombstoned.map { it.value })
+        assertEquals(1, result.pendingGcPairs.size, "the [3,5) remainder must be buffered for GC")
```
longhoang-lewy, line 258: **Minor — assert identity, not count.**

`assertEquals(1, result.pendingGcPairs.size)` passes even if `split()` starts buffering the *wrong side*, which is precisely the subtlety that makes the register-before-unregister ordering in `executeRestore` load-bearing.

```kotlin
assertEquals(listOf("5678"), result.pendingGcPairs.map { it.child.value })
```

Worth double-checking the message text while you're here: `CrdtTreeNode.split` registers only the **right** piece when `removedAt != null` (`CrdtTree.kt:2054-2059`), so the buffered pair may be the isolated target itself rather than "the [3,5) remainder" the message names — the `[3,5)` piece was already registered by the original delete. The counts coincide either way, which is why the label discrepancy is invisible today. Same for the two-boundary case at `:293-297`.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConvergenceTest.kt
```diff
@@ -0,0 +1,398 @@
...
+            length = 4,
+            value = "3456",
+        )
+
+        val (pairs, diff) = tree.retombstone(listOf(span), issueTime())
+
+        assertEquals(1, pairs.size)
+        assertEquals("3456", pairs.single().child.value)
+        assertTrue(
+            diff != DataSize(0, 0),
```
longhoang-lewy, line 335: **Minor — `diff != DataSize(0, 0)` is too weak for a case whose subject is DocSize exactness.**

Any wrong non-zero value passes, including one that double-counts. Since this is the test that pins the live-split accounting, assert the exact figure:

```kotlin
assertEquals(DataSize(0, 2 * PER_NODE_META), diff)   // two live splits
```

or, if the per-node metadata constant is awkward to reach, register `pairs` into a `CrdtRoot` and assert `docSize.live + docSize.gc` equals that of an equivalently-segmented freshly-built tree — which also pins the invariant that segmentation is size-neutral overall.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
thiennguyen-naver: Fixed. `restore`'s span loop now runs inside `try { … } finally { pairs = drainPendingGcPairs() }`, so the buffer is drained on the throw path too (the server companion's `defer`). Pinned by `TreeRestoreConvergenceTest."restore drains pending GC pairs even when a later span throws"` — a `[5,15)` span with `value = null` over a tombstoned `[3,9)` run: the isolate at 5 buffers a pair, the gap past 9 trips `requireNotNull(span.value)`, and the test asserts the tree-level buffer is empty afterwards. Mutation-verified: against the pre-fix `restore` it fails with "the throw path must drain the buffer".

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
thiennguyen-naver: Fixed — two tests, one per `addDataSizes` site. Tree level: `TreeRestoreConvergenceTest."restore isolates a live straddling piece and reports the split overhead"` (your snippet, with the exact figure: `diff == DataSize(0, 2 * TIME_TICKET_SIZE)`, nothing untombstoned/recreated/pending, XML unchanged, live pieces `["012", "3456", "789"]`). Op level: `TreeEditOperationReverseTest."restore over a live straddling piece adds the split overhead to docSize live"` — `docSize.live` grows by exactly that diff, `gc` untouched, `garbageLength == 0`. Mutation-verified: dropping the `restored.diff` accumulation in `executeRestore` fails the op-level test (`meta=168` expected vs `120`).

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
thiennguyen-naver: Added: `require(pieceStart <= from && from < to && to <= pieceEnd)` with the range and piece id in the message; KDoc now says the precondition is enforced and why (an escaping range would hand back a node covering content outside the span, which `restore` would revive). Existing isolate cases (exact/left/right/both/live) cover the valid paths.

yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeEditOperation.kt
thiennguyen-naver: Changed to named access: `val restored = tree.restore(toRestore)` then `restored.pendingGcPairs` / `restored.untombstoned` / `restored.diff` / `restored.recreated` (and in the `changed` predicate). No positional binding left on this path.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConcurrentTest.kt
thiennguyen-naver: Agreed — two local undos before a single `crossSync` cannot interact, so the old test only compared a deterministic computation with itself. Rewritten as the mirror of the interleaved case via a shared `undoInterleaved(first, second)` helper: the order test now runs `undoInterleaved(d2, d1)` (d2's undo synced first, so d1 restores around d2's segmentation), asserts convergence + initial content, and then asserts `identitySequence` equality against both the forward-interleaved run and the batched `undoBoth` run for the same relation — all three land on the same node segmentation for all six relations. That last equality (batched == interleaved) holds because isolate boundaries derive only from span offsets, as your "verified clean" note argues; it is now pinned rather than reasoned.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConcurrentTest.kt
thiennguyen-naver: Fixed. `deleteOverlapping` now asserts the purge is total on both replicas: `garbageLength == 0` and `getDocSize().gc == DataSize(0, 0)` right after `garbageCollect(maxVectorOf(actor1, actor2))`, with a comment on why (`> 0` would let a surviving tombstone route restore down the cheap `unremove()` path and silently stop exercising recreate + isolate). Holds for all six relations.

yorkie/src/androidTest/kotlin/dev/yorkie/document/json/JsonTreeRestoreTest.kt
thiennguyen-naver: Fixed. Before the undos: `assertEquals(0, d1.garbageLength)` / `d2` (and, right after the two local deletes, `garbageLength > 0` on both so the zero is a purge, not a no-op). Verified against Docker `yorkieteam/yorkie:0.7.15` on emulator-5554 — the five settle rounds do drive the min-synced vector far enough; the assertion is green, not aspirational.

yorkie/src/test/kotlin/dev/yorkie/document/CounterWrapConvergenceTest.kt
thiennguyen-naver: Added `JsonCounterWrapTest.test_out_of_int32_long_increase_wraps_identically_over_the_wire` (androidTest, `withTwoClientsAndDocuments(syncMode = Manual)`): `setNewCounter("age", 1)`, sync both, `increase(5_000_000_000L)` on c1, sync both, `705032705` asserted on **both** documents plus `d1.toJson() == d2.toJson()`. Green against server 0.7.15 — the Long delta survives `IncreaseOperation` → protobuf → server → peer untruncated. The JVM `CounterWrapConvergenceTest` stays as the fast arithmetic pin. Agreed on the congruence argument; the class KDoc points at it.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConvergenceTest.kt
thiennguyen-naver: Fixed, and you were right about the label. `CrdtTreeNode.split` buffers the **right** piece of a removed split, so in the left-boundary case the buffered pair is the isolated target itself (`"5678"`), not `[3,5)` — which keeps the original delete's pair. The three cases now assert identity: left → `listOf("5678")` plus `assertSame(untombstoned.single(), pendingGcPairs.single().child)` (the reason the register-before-unregister order is load-bearing); right → `listOf("78")`; both → `listOf("56", "78")` (the target, narrowed by the second split, then `[7,9)`). Messages and the leading comment corrected to match.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConvergenceTest.kt
thiennguyen-naver: Fixed: `assertEquals(DataSize(0, 2 * TIME_TICKET_SIZE), diff)` — a live text split adds exactly one node's ticket overhead (`splitText`: left + right − original = `(0, TIME_TICKET_SIZE)`), two splits here. Same exact figure in the new restore-side live-isolate test.

thiennguyen-naver:
Review 5036530353 — all items addressed in 1198e94f.

**Blocker (stale base):** already resolved on the branch before this round — #360 merged into `develop` as `21b4c1ac`, this branch is rebased onto it (merge-base `21b4c1ac`) and the PR now targets `develop`. Verified on HEAD: `coerceIn` 0 refs in `CrdtTree.kt`/`TreeEditOperation.kt`, `traverseAllPreorder` present, the four `isNullOrEmpty` encoder guards present; `crossSync` has `overWire`. The GitHub diff is now the real v0.7.15 delta.

**Important**
- I1 — `restore` drains `pendingGcPairs` in a `finally` (server `defer` parity); leak-on-throw test, mutation-verified.
- I2 — live-straddler isolate inside `restore` tested at tree level (exact `DataSize(0, 2 * TIME_TICKET_SIZE)`) and op level (`docSize.live` delta); mutation-verified against a dropped `restored.diff`.
- I3 — order test rewritten as reverse-interleaved vs forward-interleaved vs batched `identitySequence` equality (holds for all six relations).
- I4 — `deleteOverlapping` asserts `garbageLength == 0` and `docSize.gc == 0` on both replicas after the purge.
- I5 — instrumented after-GC case asserts `garbageLength > 0` before settling and `== 0` before the undos; green vs server 0.7.15.
- I6 — new instrumented `JsonCounterWrapTest`: `5_000_000_000L` relayed through server 0.7.15 reads `705032705` on both clients.

**Minor**
- Decoder now rejects `isText && length == 0` (same boundary as the existing `length < 0` rule; test added). Note for the record: like the negative-length rejection, this is an Android-side decode rejection of a payload no SDK produces — JS/Go would apply the one-character retombstone. Happy to fold a `length >= 1` requirement into the upstream draft so all three close it together.
- `@Ignore`d known limitations documented in `CrdtTree.restore`'s KDoc.
- Relation loops collect failures via JUnit's `ErrorCollector` (`forEachRelation`) instead of stopping at the first.
- `isolateTextRange` precondition enforced with `require`; `TreeRestoreResult` accessed by name in `executeRestore`; pending-pair identity and exact live-split diff asserted.

Verification: unit lane 758 tests, 0 failures, 3 skipped (2 upstream it.skip twins + 1 pre-existing), `lintKotlin` clean, connected `JsonTreeRestoreTest` + `JsonCounterWrapTest` on emulator-5554 vs `yorkieteam/yorkie:0.7.15`.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConcurrentTest.kt
longhoang-lewy: Verified addressed at `7cd5201` — resolving.

The fix landed as an addition rather than an edit: `converges when one replica syncs its undo before the other undoes` runs the whole `overlapRelations` matrix as `d1.undo -> crossSync -> d2.undo -> crossSync`. That is the genuinely order-sensitive shape, and it closes the failure scenario I described — d2's restore now runs against d1's already-restored live segmentation instead of a fully purged run, so a regression in that path can no longer ship green. The new comment above it states that scoping explicitly.

Keeping the batched matrix test was the right call; as I said originally, it is still a worthwhile per-relation convergence and initial-content pin.

One residual nit, not worth blocking on: the batched test is still named `...regardless of undo order` and its added `identitySequence` comparison is between two runs that never interact, so that particular claim remains vacuous in isolation. The invariant is genuinely covered now by the interleaved test next to it — consider renaming the batched one to what it actually pins (per-relation convergence) so the suite does not read as asserting order-independence twice, once for real and once not.

thiennguyen-naver:
CI on `1198e94f` (run 34832086581) failed a single instrumented test out of 366: `DocPresenceTest.test_receive_document_events_according_to_presence_changes`, a 15 s timeout on all three `RetryRule` attempts. Every test this PR adds or touches passed (`JsonTreeRestoreTest` ×2, `JsonCounterWrapTest`), presence code is untouched by this branch, and the same test was green on the previous head `7cd52014` 30 min earlier.

**Root cause (test-side race, pre-existing).** `withTwoClientsAndDocuments` attaches d1 then d2 without waiting for d1's watch stream, and the test relied on `delay(1000)` before d2's presence put. When d1's Watch RPC comes up after c2 attached, c2 arrives in the stream's `Initialization` frame — which sets `onlineClients` and emits `Initialized` (a `MyPresence` event), never `Others.Watched` — and a put applied before d1 holds c2's initial presence folds into a late `Watched` with no `PresenceChanged`. The test filters `Others` and waits for two events before detaching, so it starves. Reproduced deterministically by attaching c2 before c1's stream connects: `[Watched]` only, then the timeout.

**Fix (`001874af`, test only):** attach c1 inside `awaitWatchConnected(d1)` before c2 attaches, and wait for the `Watched` event itself instead of the fixed delay — the same shape `test_presence_changed_event_with_multiple_changes` already uses. 3/3 local runs of the class green against server 0.7.15.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConcurrentTest.kt
```diff
@@ -0,0 +1,466 @@
...
+        d1.setActor(actor1)
+        d2.setActor(actor2)
+
+        d1.updateAsync { root, _ ->
+            root.setNewTree(
+                "t",
+                element("doc") { element("p") { text { "0123456789" } } },
+            )
+        }.await()
+        crossSync(d1, d2)
```
longhoang-lewy: **(Important) The new convergence matrix never goes over the wire.**

Every `crossSync` in this file uses the default `overWire = false`, so relayed operations skip `toPBOperation()` / `toOperations()` entirely. `Helper.kt`'s own KDoc calls `overWire` "the only way a unit test exercises the `restore_spans` encode/decode path under convergence", and right now exactly one test uses it — `TreeRestoreConvergenceTest.kt:164`, which predates this PR.

That matters here specifically because this PR changes *both* sides of that path: what `restore` emits (split-aware spans that peers must replay to converge on segmentation) and the decode-side validation in `toTreeRestoreSpan`. So the 12-case matrix — the main evidence for the change — proves in-memory convergence but not wire convergence.

The instrumented `JsonTreeRestoreTest` does cover the real wire, but only two cases, and it asserts XML equality rather than the node-identity sequence that is the actual point of the fix.

Cheapest fix: thread an `overWire` parameter through `seed`/`deleteOverlapping`/`undoBoth` and run at least one relation (`contained_by` would do) with `overWire = true`.

yorkie/src/androidTest/kotlin/dev/yorkie/document/presence/DocPresenceTest.kt
```diff
@@ -724,6 +724,7 @@ class DocPresenceTest {
     fun test_receive_document_events_according_to_presence_changes() {
         withTwoClientsAndDocuments(
             syncMode = Realtime,
+            attachDocuments = false,
```
longhoang-lewy, line 727: **(Important — scope) This is unrelated to the v0.7.15 sync, and the PR description doesn't mention it.**

The fix itself is well-reasoned and looks correct to me: waiting for d1's watch stream before c2 attaches, and waiting for the `Watched` event instead of `delay(1000)`, addresses a real ordering race — a peer already watching when the stream connects arrives in the `Initialization` frame and never produces an `Others.Watched` event. Good root-cause analysis in the commit message.

The problem is discoverability. The Summary says the diff is "the v0.7.15 delta only", the Changes section never lists `DocPresenceTest.kt`, and the instrumented test-plan line names only `JsonTreeRestoreTest` and `JsonCounterWrapTest`. Anyone bisecting a presence-test regression later won't think to look in a tree-restore sync PR.

Either split it into its own PR (it stands on its own and needs no 0.7.15 context) or add it to Changes and the test plan.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
```diff
@@ -1389,74 +1434,133 @@ internal data class CrdtTree(
...
+     * covering content outside the span (which [restore] would then revive).
+     */
+    private fun isolateTextRange(
+        piece: CrdtTreeNode,
+        from: Int,
+        to: Int,
+    ): Pair<CrdtTreeNode, DataSize> {
+        val pieceStart = piece.id.offset
+        val pieceEnd = pieceStart + piece.value.length
+        require(pieceStart <= from && from < to && to <= pieceEnd) {
```
longhoang-lewy: **(Minor) `require` surfaces an unchecked `IllegalArgumentException` out of a document-apply path.**

The precondition itself is right, and I checked that it is currently unreachable from a decoded wire payload: `findPiecesOverlapping` only returns pieces that genuinely overlap `[start, end)`, which makes `max(pieceStart, start) < min(pieceEnd, end)` hold at the `retombstone` call site, and `restore`'s cursor invariant gives the same at its site. So this is hardening, not a live bug.

But when it does fire it throws `IllegalArgumentException`, not `YorkieException(ErrInvalidArgument)` — and `restore drains pending GC pairs even when a later span throws` now pins that behaviour (via `requireNotNull(span.value)` in `recreateFromSpan`, same story). The iOS twin throws `YorkieError(code: .errInvalidArgument)` here, and this file's own converter uses `YorkieException`. An unchecked exception escaping into application code is worse than a typed one for an SDK.

Suggest making both throw `YorkieException(ErrInvalidArgument, ...)` and updating the test's `assertFailsWith`.

yorkie/src/main/kotlin/dev/yorkie/api/OperationConverter.kt
```diff
@@ -453,7 +457,7 @@ private fun PbTreeRestoreSpan.toTreeRestoreSpan(): TreeRestoreSpan {
     val malformed = !hasId() || !id.hasCreatedAt() ||
         anchors.any { (present, anchor) -> present && !anchor.hasCreatedAt() } ||
         attributesMap.values.any { !it.hasUpdatedAt() } ||
-        length < 0 || (isText && value.length != length)
+        length < 0 || (isText && (length == 0 || value.length != length))
```
longhoang-lewy: **(Minor — cross-SDK) This rejection is Android-only.**

The reasoning is sound — `retombstone`'s `maxOf(span.length, 1)` clamp really would widen a zero-length text span into re-removing a character the sender never deleted — and it's unreachable from today's capture.

But iOS's `fromTreeRestoreSpan` validates no length at all (I checked `Converter.swift`), and `TreeRestoreConverterTest`'s own neighbouring case notes that "JS validates no span length at all". So an Android client now hard-rejects a payload that JS and iOS peers accept and process. Given the rule that the wire contract stays consistent across all four SDKs, a decode-side divergence is the kind of thing that bites much later, in a mixed-client session, as an Android-only failure to apply a remote change.

Two options: note in the PR that this is deliberate Android-side hardening ahead of the others, or — better — fix the root cause upstream by removing the `max(span.length, 1)` clamp in `yorkie-js-sdk`, so all four SDKs agree and the guard becomes unnecessary.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConcurrentTest.kt
```diff
@@ -0,0 +1,466 @@
...
+    /**
+     * Runs [block] once per relation in [overlapRelations], collecting each
+     * relation's failure instead of stopping at the first, so one red
+     * relation cannot hide the others.
+     */
+    private suspend fun forEachRelation(
+        block: suspend (label: String, r1: Pair<Int, Int>, r2: Pair<Int, Int>) -> Unit,
+    ) {
+        for ((label, r1, r2) in overlapRelations) {
+            runCatching { block(label, r1, r2) }
```
longhoang-lewy: **(Nit) `runCatching` catches `Throwable`, including `CancellationException`.**

Inside `runTest`, a cancellation would be swallowed and recorded as a relation failure rather than propagating. Narrowing to `catch (e: AssertionError)`, or rethrowing `CancellationException`, keeps the per-relation collection without that edge.

**(Nit, same file)** `overlapRelations` above hard-codes the same six ranges that the twelve explicit `@Test` methods spell out individually — the two can silently drift. Driving the per-relation tests off the same list (or at least a shared `val` per relation) would keep them in step.

longhoang-lewy (commented):
Re-review after the rebase and the round of follow-ups. I pulled the branch into a worktree, cross-checked it against the shipped iOS 0.7.15 port of the same upstream change, and ran the suite locally.

**The core change is sound.** `restore`/`retombstone` are now split-aware via `isolateTextRange`, and the implementation is essentially line-for-line with the iOS port: same cursor walk, same split order, same `maxOf(span.length, 1)` clamp, same register-before-unregister ordering in `executeRestore`.

### Claims I verified rather than trusted

- **Test plan** — reproduced exactly: 758 tests, 0 failures, 3 skipped; `lintKotlin` clean.
- **GC accounting on the new path.** This was my main worry going in: `registerPendingGcPair` registers with `gcOnlySize = <split diff>`, but `unregisterGCPair` credits the node's *full* `dataSize` — so the new register-pending-then-unregister of a split-born target looked asymmetric. I probed it directly (partial restore of a tombstoned run, then purge-all): `docSize.gc` returns to exactly `DataSize(0, 0)`, and `live.data` matches a freshly-built equivalent document. The extra meta is two genuine additional text nodes, not a leak. **No bug** — the ordering comment in `executeRestore` is doing what it says.
- **The counter no-code determination is correct.** `CrdtCounter.increase` is `value.toInt() + primitiveValue.toInt()`; Kotlin `Int + Int` wraps two's-complement and `Long.toInt()` keeps the low 32 bits, so it computes `wrap(value + delta)` — congruent with the fixed JS `bigintToInt32(BigInt(value) + delta)`. The pinned values check out (`1 + 5_000_000_000 → 705032705`, `0 + 2^31 → Int.MIN_VALUE`, `-1 + 2^31 → Int.MAX_VALUE`).
- **The `changed` gate.** The added `diff != DataSize(0, 0)` genuinely closes the gap it claims to: a removed straddler's split always yields a removed target (so `untombstoned` is non-empty), and a live split always yields non-zero meta — `TreeEditOperationReverseTest` pins it at `2 * TIME_TICKET_SIZE`.
- **The stale-base blocker is resolved.** The #360 fixes (`coerceIn` removal, `traverseAllPreorder`, the `isNullOrEmpty` encoder guards, `crossSync(overWire)`) are all present on HEAD.

The test suite is genuinely strong — the identity-sequence assertions, the total-purge precondition, and the exact pending-pair identity checks are the right things to pin.

### Findings

Two worth acting on before merge (neither is a correctness bug), plus three small ones — see the inline comments.

Nothing here blocks on correctness; my earlier `CHANGES_REQUESTED` is satisfied. Leaving this as a comment rather than an approval only so the merge call stays with you.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConcurrentTest.kt
thiennguyen-naver: Fixed. `overWire` is threaded through `seed`/`deleteOverlapping`/`undoBoth`/`redoBoth`/`undoInterleaved`, and a new `converges over the wire on undo and redo for every relation` runs the whole matrix (all six relations, undo **and** redo) with every relayed operation routed through `toPBOperation()`/`toOperations()`, asserting node-identity convergence via `assertConverged`. `JsonTreeRestoreTest` now asserts the postorder identity sequence alongside XML in both cases (green against server 0.7.15).

yorkie/src/androidTest/kotlin/dev/yorkie/document/presence/DocPresenceTest.kt
thiennguyen-naver: Kept in this PR (it is what makes this PR's CI deterministic) and made discoverable: the Summary no longer claims a pure v0.7.15 delta, Changes has a dedicated bullet for `DocPresenceTest.test_receive_document_events_according_to_presence_changes` with the root cause and the `001874af` reference, the re-review section lists it, and the instrumented test-plan command now includes `DocPresenceTest`.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
thiennguyen-naver: Fixed at both sites: the `isolateTextRange` precondition and `recreateFromSpan`'s missing text value now throw `YorkieException(ErrInvalidArgument, …)` — the iOS twin's error and the convention `ElementRht`/`RgaTreeList` already follow inside `crdt/`. The drain-on-throw test asserts `YorkieException` with code `ErrInvalidArgument`. The wider `check`/`require` sites in `IndexTree` (#360 thread 3860045091) stay out of this PR.

yorkie/src/main/kotlin/dev/yorkie/api/OperationConverter.kt
thiennguyen-naver: Reverted. The `isText && length == 0` rejection is gone and the decoder is back to the #360 shape; the case is pinned as a parity test (`tolerates a zero-length text span with an empty value`, length 0 and empty value preserved). One point beyond cross-SDK consistency tipped it: a hostile peer can already delete that single character with a legitimate edit, so the guard protected nothing and only added an Android-only failure to apply a remote change. Upstream draft (Draft 5 in our review-drafts list, to be filed with the others) proposes all SDKs drop the shared `max(span.length, 1)` clamp — a zero-length span then becomes a no-op in both directions, as `restore` already treats it — or reject `length <= 0` everywhere at once.

yorkie/src/test/kotlin/dev/yorkie/document/TreeRestoreConcurrentTest.kt
thiennguyen-naver: Both fixed. `forEachRelation` now uses `try/catch`: `CancellationException` is rethrown, everything else is collected per relation. The six relations are one `private val` each (`containedBy`, `contains`, `overlapStart`, `overlapEnd`, `identical`, `adjacent`) of a small `Relation(label, r1, r2)` data class; `overlapRelations` lists those vals and the twelve explicit tests take the same vals, so the two cannot drift.

thiennguyen-naver:
Re-review 5205146566 — all five items addressed in `e9074b5f` + `8f2c7709`.

- **R1** — the six-relation matrix now also runs undo+redo **over the wire** (every relayed op through the protobuf converters) with node-identity convergence; `JsonTreeRestoreTest` asserts the identity sequence, not just XML.
- **R2** — kept in this PR, documented: Summary, a dedicated Changes bullet with the root cause, the re-review section, and the instrumented test-plan command all name the `DocPresenceTest` fix.
- **R3** — `isolateTextRange` precondition and `recreateFromSpan` missing value throw `YorkieException(ErrInvalidArgument)`; test asserts the type and code.
- **R4** — Android-only `length == 0` rejection **reverted** (JS/iOS parity; a peer can already delete that character legitimately). Parity test added; upstream draft proposes dropping the shared `max(length, 1)` clamp in all SDKs together.
- **Nits** — `CancellationException` propagates from the per-relation collector; one `Relation` val per relation feeds both the list and the explicit tests.

Verification: unit lane 759 tests, 0 failures, 3 skipped (2 upstream it.skip twins + 1 pre-existing), `lintKotlin` clean, connected `JsonTreeRestoreTest` + `DocPresenceTest` 16/16 on emulator-5554 vs `yorkieteam/yorkie:0.7.15`. PR body refreshed.

longhoang-lewy (approved):
Approving — all five findings from #pullrequestreview-5205146566 are resolved, and CI is green on `8f2c770` (Checks, instrumented + unit tests on API 33, Codecov).

I re-ran the suite locally on the new head: **759 tests, 0 failures, 3 skipped** (+1 over the previous head — the new over-the-wire case, confirmed executed rather than skipped), `lintKotlin` clean.

**What changed since my review**

1. **Wire coverage** — you went further than I asked: `converges over the wire on undo and redo for every relation` runs all six relations through the protobuf converters for undo *and* redo, and `JsonTreeRestoreTest` now asserts the node-identity sequence rather than just XML. That second one matters more than it looks — identical rendered text was exactly what could have hidden a segmentation divergence in the instrumented tests.
2. **Scope** — the `DocPresenceTest` race fix is now documented under Changes and in the test plan, flagged as unrelated to v0.7.15. That is all I wanted; no need to split it.
3. **Typed errors** — both the `isolateTextRange` precondition and `recreateFromSpan`'s missing-value path now throw `YorkieException(ErrInvalidArgument)`, matching the iOS twin, and the drain-on-throw test pins the code.
4. **Zero-length span parity** — removing the guard rather than tightening it is the right call. It keeps Android from rejecting a payload its peers accept, and puts the real fix (`retombstone`'s `maxOf(span.length, 1)` clamp) upstream where all four SDKs get it at once. The converter test reworked into a decode-and-round-trip parity pin is a better test than the rejection assertion it replaced.
5. **Nits** — `CancellationException` rethrown, and the `Relation` data class removes the drift risk between the matrix and the explicit per-relation tests.

**Re-verified on the new head**

I re-checked the precondition-reachability claim I made last round, this time all the way down rather than by inference: `splitText` returns `null` at both boundaries and when the right half would be empty, and the `value` setter keeps `visibleSize == value.length` for text nodes — so no zero-length text piece can exist, and `isolateTextRange`'s precondition is unreachable from a decoded payload. That still holds now that zero-length spans decode successfully: `restore`'s loop no-ops at `length == 0`, and `retombstone`'s clamp keeps `from < to` via `findPiecesOverlapping`'s overlap guard.

The GC-accounting result from last round is unchanged and still the thing I'd point at if this ever regresses: a partial restore of a tombstoned run followed by purge-all returns `docSize.gc` to exactly `DataSize(0, 0)`, so the register-pending-before-unregister ordering in `executeRestore` is doing what its comment claims.

**Two things for later, neither blocking**

- `node = requireNotNull(right)` in `isolateTextRange` is still an unchecked throw where the iOS twin uses a typed error. Provably unreachable given the precondition above, so purely cosmetic — worth folding into whatever touches this next.
- The `max(span.length, 1)` clamp in `yorkie-js-sdk` is the actual root cause behind finding 4. Worth an upstream issue so the Android/iOS/Go decoders can eventually stop tolerating a span shape that should never be emitted.

Nice work on the test suite in particular — the total-purge precondition, the per-relation failure collection, and the exact pending-pair identity assertions are the right invariants to have pinned.

PR #362, opened 2026-08-19
[Android] Sync-up Yorkie JS SDK v0.7.16

thiennguyen-naver:
> **RTCOLLABPLATFORM-756**: [[Android] Sync-up Yorkie JS SDK v0.7.16](https://jira.navercorp.com/browse/RTCOLLABPLATFORM-756)

## Summary
- Ports all three yorkie-js-sdk v0.7.16 Tree fixes (#1318, #1317, #1319) — duplicate-node-ID resilience, merge-bound insert stamping + style-range interloper filtering, and identity-fresh undo copies with wire-carried element-split tickets
- One additive proto change: `Operation.TreeEdit.split_tickets = 11` (empty-field payloads replay via the old reconstruction)
- Pins the local Yorkie server image 0.7.15 → 0.7.16 (server companions yorkie#1927/#1928/#1929)

## Changes
- **Duplicate-ID resilience (JS `2ed28322`)**: `CrdtTree.registerNode` keeps a live node over a tombstone claiming the same ID; guarded purge removes only the purged node's own map entry; new `dropDuplicateContents` drops edit contents reusing an ID from another change (own-change collisions kept); constructor detects duplicates on snapshot rebuild and re-registers; `IndexTreeNode.splitText` refuses out-of-range offsets with `YorkieException(ErrInvalidArgument)`; accepted `insertedContentSize` threaded so a dropped copy never widens the reverse operation
- **Merge-bound stamping + interloper filter (JS `1c033ff5`)**: inserts declared inside a merged-away parent carry `mergedFrom`/`mergedAt`; new `mergedAnchorInterloperGuard` + `styleSkipPredicate` shared by `style`/`removeStyle` (fail-open on any merge stamp); `cloneElement`/`cloneText` carry the stamps
- **Split tickets + undo identity (JS `4ec66cc0`)**: `JsonTree.editInternal` records issued split tickets; `TreeEditOperation` consumes carried tickets then falls back to reconstruction; `Document.executeUndoRedo` reserves the ticket range `buildFreshNodes` mints so later ops in the same undo change never repeat a minted ID (Android-shape equivalent of JS `reissueContentIDs`); the snapshot rebuild refuses a splitting edit (`check(splitLevel == 0)`, upstream `ErrRefused`); `toReverseOperation` skips the reverse when its range would run past the post-edit tree (upstream no-op guard)
- Tests: 33 new JVM cases across 6 suites (duplicate-ID, moved-anchor style cross-sync ×7, split-ticket wire, undo identity, split guard) + 1 instrumented two-client case (`JsonTreeStyleMovedAnchorTest`) against the real 0.7.16 server

## Intentionally not ported
- **`reissueContentIDs`** (JS `4ec66cc0`): every Android reverse sets `contents = null` and rebuilds from `TreeNode` snapshots (`buildFreshNodes`), which carry no `insPrevID`/`insNextID`/`merged*`; JS needs the explicit re-identification because its reverse deep-copies live nodes. Android's counterpart is the `Document.executeUndoRedo` ticket reservation plus the `check(splitLevel == 0)` guard at the rebuild site
- **`getContentSize()`** (JS `2ed28322`): `Document.reconcileHistoryEdits` reads `opInfo.nodes?.size`, sourced from `aliveContents`, which already excludes a dropped copy

## Test plan
- [ ] `docker compose -f docker/docker-compose.yml up --build -d && ./scripts/config-yorkie-local-server.sh` (server 0.7.16)
- [ ] `./gradlew yorkie:connectedDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=dev.yorkie.document.json.JsonTreeStyleMovedAnchorTest` — both replicas converge, inserted `<x>` carries no bold

> Items run automatically by CI (lint, unit tests, coverage) are excluded.


yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeEditOperation.kt
```diff
@@ -428,12 +438,13 @@ internal data class TreeEditOperation(
...
-        // contributes its [paddedSize]: text length for text nodes, or
-        // `2 + sum(children.paddedSize)` for elements (open + close + body).
-        val insertedSpan = editContents?.sumOf { it.paddedSize } ?: 0
+        // Document-index span occupied by what was actually inserted. Read
+        // from CrdtTree.edit's accepted-content measurement rather than
+        // recomputed from contents here: a copy dropped as a cross-change ID
+        // reuse (CrdtTree.dropDuplicateContents, port 2ed28322) contributes
+        // nothing, so this reverse never widens past what was truly
+        // inserted (redo would otherwise delete a neighbour).
+        val insertedSpan = insertedContentSize
```
longhoang-lewy, line 447: This silently changes `insertedSpan` from a **post-insert** to a **pre-insert** measurement, and there's no guard behind it.

The old code evaluated `editContents.sumOf { it.paddedSize }` here, i.e. *after* `tree.edit` had already spliced those exact node objects in. When `fromParent.isRemoved`, `CrdtTree.edit` tombstones each one, and `CrdtTreeNode.remove` calls `updateAncestorSize(-paddedSize())` — so an element content's `visibleSize` collapsed to 0 and this read back near-zero. `result.insertedContentSize` is measured while still detached, so it now reads the **full** span.

That matches upstream's semantics, but upstream pairs it with a guard we don't have (`tree_edit_operation.ts`):

```ts
const maxNeededIdx = preEditFromIdx + insertedContentSize;
if (maxNeededIdx > tree.getSize()) return undefined;  // edit was a no-op
```

We go straight to `reverseToIndex = fromIndex + insertedSpan`. Reachable shape: one edit that both deletes a range (so `opInfos` is non-empty) *and* inserts under a removed `fromParent` (so `spansComplete` is false → this copy-reinsert branch). The reverse then gets `undoToOffset` past the live tree — and per the comment at the top of `execute()`, offsets are deliberately **not** clamped, so `indexRangeToPosRange` is meant to fail loudly. Undo would throw rather than no-op.

I couldn't construct a failing test for this, so treat it as unverified — but the guard is three lines and restores what the old post-insert measurement was doing incidentally.

yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeEditOperation.kt
```diff
@@ -573,7 +584,14 @@ internal data class TreeEditOperation(
     private fun issueTimeTicket(executedAt: TimeTicket): () -> TimeTicket {
         var delimiter = executedAt.delimiter
         contents?.let { delimiter += it.size.toUInt() }
-        return { TimeTicket(executedAt.lamport, ++delimiter, executedAt.actorID) }
+        var issued = 0
```
longhoang-lewy, line 587: Upstream refuses this case explicitly in `reissueContentIDs`:

```ts
if (this.splitLevel !== 0) {
  throw new YorkieError(Code.ErrRefused, `cannot reissue content ids on a splitting edit`);
}
```

because `buildFreshNodes` consumes `executedAt.delimiter + 1 … + countNodes`, while the fallback branch just below starts at `executedAt.delimiter + contents.size + 1` — the two ranges overlap as soon as a content node has descendants.

We hold that invariant only implicitly: every `removedNodeSnapshots`-carrying reverse is built with `splitLevel = 0`, and the `redoSplitLevel`/`mergeLevel` reverses set `contents = null`. Nothing catches a future change to `toReverseOperation` that breaks it, and the failure mode is a silent duplicate ID — exactly what #1318/#1319 exist to prevent. Worth a `require`/`check` mirroring upstream.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
```diff
@@ -1104,10 +1156,66 @@ internal data class CrdtTree(
...
+            return
+        }
         nodeMapByID[node.id] = node
     }
 
+    /**
+     * Filters [contents] before they are spliced into the tree at
+     * [editedAt]: a content subtree whose ID was already claimed by a
+     * DIFFERENT change (an earlier change, another actor, or an ID this
+     * same edit's own split is about to create) is dropped as a whole
```
longhoang-lewy, line 1185: This KDoc contradicts itself. The first paragraph lists *"an ID this same edit's own split is about to create"* among what gets dropped; the next sentence says a subtree whose reused ID belongs to THIS SAME change/actor is kept.

The code keeps them — `lamport == editedAt.lamport && actorID == editedAt.actorID` → `return@traverseAll` — and `CrdtTreeDuplicateIdTest.'dropDuplicateContents keeps a subtree whose id collides with this edit's own change'` pins that. The third clause of the parenthetical should just come out.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
```diff
@@ -1104,10 +1156,66 @@ internal data class CrdtTree(
...
+     * Registers [node] under its [CrdtTreeNode.id] in [nodeMapByID], keeping
+     * a live node over a tombstone when two nodes claim the same ID (a
+     * document whose history re-inserted a deleted copy under its original
+     * ID, port 2ed28322). A refused node stays reachable via tree traversal,
+     * just not via lookup. Same-state pairs (both live or both removed)
+     * stay last-registered-wins — element-split delimiter IDs legitimately
+     * collide.
+     */
     fun registerNode(node: CrdtTreeNode) {
+        val entry = nodeMapByID.floorEntry(node.id)
```
longhoang-lewy, line 1169: nit: `floorEntry(node.id)` + `entry.key == node.id` is just `get(node.id)` here — three sites in this file (`registerNode`, `delete`, `dropDuplicateContents`). `nodeMapByID` is a `java.util.TreeMap`, not JS's `LLRBTree`, which has no `get`, so the idiom didn't survive the port with a purpose.

I checked it's behaviourally identical: `CrdtTreeNodeID.compareTo` and `TimeTicket.compareTo` both agree with their data-class `equals` (same fields, different ordering), so `floorEntry(k).key == k` can't disagree with `containsKey(k)`. Just an extra comparison and an odd read in Kotlin — `nodeMapByID[node.id]` says the same thing.

longhoang-lewy (commented):
Reviewed this against the three JS commits (`2ed28322` #1318, `1c033ff5` #1317, `4ec66cc0` #1319). The port is faithful — proto field 11 matches upstream exactly (number/type/cardinality), the `registerNode` / guarded-purge / `dropDuplicateContents` / constructor-recovery set matches #1318, the `mergedAnchorInterloperGuard` + stamping set matches #1317 (the inlined `isRemovalUnknown` is the exact negation of upstream's `ticketKnown`), and `countNodes()` matches `buildFreshNodes` one-ticket-per-node.

Two upstream pieces are correctly *not* ported, worth recording so nobody chases them:

- **`reissueContentIDs`** — every reverse here sets `contents = null` and rebuilds from `TreeNode` snapshots, which structurally carry no `insPrevID`/`insNextID`/`merged*`. JS needed the explicit field-clearing only because its reverse deep-copies live nodes.
- **`getContentSize()`** — `Document.reconcileHistoryEdits` reads `opInfo.nodes?.size`, sourced from `aliveContents`, which already excludes a dropped copy.

Four inline notes below; only the first is substantive. Note CI hasn't validated this yet — `Run instrumented tests and unit tests` is still in progress, so the headline `JsonTreeStyleMovedAnchorTest` result is the gate.

yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeEditOperation.kt
thiennguyen-naver: Guard ported: `if (fromIndex + insertedSpan > tree.size) return null` — the shape `toSplitReverseOperation` already uses; `toReverseOperation` is now nullable and the call site's `listOfNotNull` handles it.

I tried to reach the shape you describe and could not, for a reason upstream of the measurement: a Local op whose `fromPos` resolves inside a tombstoned parent inserts nothing alive, and `canDelete` returns false for tombstone-known nodes, so it deletes nothing either — `opInfos` is empty and `execute()` never calls `toReverseOperation` (the `opInfos.isNotEmpty()` gate). A range that crosses out of the removed parent into a live one is a merge and takes the `mergeLevel` branch (`contents = null`). So the guard is parity plus a contract statement for the pre-insert measurement, not a fix for a path I can pin; no test added for that reason.

yorkie/src/main/kotlin/dev/yorkie/document/operation/TreeEditOperation.kt
thiennguyen-naver: Added `check(splitLevel == 0)` at the `buildFreshNodes` branch in `execute()` — the Android analog of `reissueContentIDs`, and the point where the two delimiter ranges would start to overlap. `IllegalStateException` rather than a `YorkieException`: `Code` has no `ErrRefused`, and this is an internal invariant, never wire data (`removedNodeSnapshots` is not serialized). Pinned in `TreeEditOperationReverseTest`: a snapshot-carrying undo op with `splitLevel = 1` throws and the tree is untouched.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
thiennguyen-naver: The clause describes a different collision than the kept one, but the wording hides it. A text piece that step-01 range resolution splits off carries the ORIGINAL text node's `createdAt` with a new offset — in `dropDuplicateContents drops content whose id this edit is about to create by splitting` the copy is `(ticket(2), 5)` against `editedAt = ticket(3)` — so by lamport/actor it belongs to another change and is dropped. Element-split delimiter IDs are minted from this edit's own ticket; that is the kept case. Removing the clause would leave that ordering pin (drop check after step 01) undocumented, so I reworded it instead, here and in the step-05 comment in `edit`.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtTree.kt
thiennguyen-naver: Taken at all three sites (`registerNode`, `dropDuplicateContents`, `delete`) — `nodeMapByID[node.id]`. `findFloorNode` keeps `floorEntry`: its `createdAt`-only match is the point there. Same equivalence check on my side: `CrdtTreeNodeID.compareTo` orders by (createdAt, offset) and `TimeTicket.compareTo` by (lamport, actorID, delimiter), the same field sets as their data-class `equals`.

thiennguyen-naver:
Addressed in cb1201fe: F1 no-op guard ported (no reachable op-level shape found — details in thread), F2 `check(splitLevel == 0)` + pin, F3 KDoc reworded (clause kept — it names the text-split case, not the same-change case), F4 `get` at all three sites. Recorded the two intentionally-unported pieces (`reissueContentIDs`, `getContentSize`) in the PR body. CI on f5f4c2ac finished green after your review (run 34943576254: instrumented + unit, Checks, Codecov).

yorkie/src/main/kotlin/dev/yorkie/document/crdt/TreeInfo.kt
```diff
@@ -58,6 +58,18 @@ internal value class TreeTextNode(override val value: String = "") : TreeNode, J
     }
 }
 
+/**
+ * Counts every node in this subtree, text and element alike — matching
+ * one ticket per node in [dev.yorkie.document.operation.TreeEditOperation]
+ * `.Companion.buildFreshNodes`, so the copy-reinsert undo path can reserve
```
longhoang-lewy, line 64: nit (`kdoc-rule.md`: class/function references use `[ClassName]`, not backticks) — this splits one reference across a bracket link and a backticked tail, so the member half doesn't resolve as a link:

```kotlin
 * one ticket per node in [dev.yorkie.document.operation.TreeEditOperation]
 * `.Companion.buildFreshNodes`, so the copy-reinsert undo path can reserve
```

KDoc will link the whole path in one go:

```kotlin
 * one ticket per node in
 * [dev.yorkie.document.operation.TreeEditOperation.Companion.buildFreshNodes],
 * so the copy-reinsert undo path can reserve
```

Non-blocking.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/TreeInfo.kt
```diff
@@ -118,4 +130,12 @@ internal data class TreeOperationResult(
      * edit was merge/split-free, mirroring [removedSpans].
      */
     val insertedSpans: List<TreeRestoreSpan> = emptyList(),
+    /**
+     * Sum of the accepted contents' `paddedSize`, measured while still
```
longhoang-lewy, line 134: nit (same `kdoc-rule.md` point) — `` `paddedSize` `` is a function/property reference, so it wants brackets: `[dev.yorkie.util.IndexTreeNode.paddedSize]`. Worth the fully-qualified form here since `TreeInfo.kt` doesn't import `IndexTreeNode`.

The neighbouring `[CrdtTree.dropDuplicateContents]` two lines down is exactly the style being asked for, which is what makes this one stand out. Non-blocking.

longhoang-lewy (approved):
Approving. All four points from review 5207333774 are addressed in `cb1201f`, and CI is now fully green — including `Run instrumented tests and unit tests (33, google_apis)` at 22m8s, which carries `JsonTreeStyleMovedAnchorTest` and the new refusal test. That job was the gate.

**Re-verified each fix:**

1. **No-op reverse guard** — `if (reverseToIndex > tree.size) return null`, return type now nullable; the single caller already wraps in `listOfNotNull`, and placement matches upstream (after the `redoSplitLevel`/`mergeLevel` early returns, before snapshot construction). Thanks for probing the reachability rather than taking my hypothesis at face value — the code comment says plainly that no op-level path reaches that shape today, which is the right way to record a parity guard.
2. **`check(splitLevel == 0)`** — correctly placed before `buildFreshNodes`, and `IllegalStateException` over a typed `YorkieException` is the right call here: `isUndoOp` is `undoFromOffset != NotAnUndoOp`, and the converter leaves that at its default for decoded remote ops, so this can only ever fire on a locally-built reverse. Good test — asserting the tree is untouched, not just that it throws.
3. **`dropDuplicateContents` KDoc** — better than what I suggested. Rather than deleting the bad clause you replaced it with the actual distinction: a text split keeps the original `createdAt` with a new offset (another change's ID, dropped), an element split mints from this edit's own ticket (same change, kept). I checked `cloneText(offset)` → `clone(offset, id.createdAt)` against `cloneElement` → `clone(0, issueTimeTicket())` — the doc is accurate, and it now explains the `…about to create by splitting` test, which uses `CrdtTreeNodeID(ticket(2), 5)`.
4. **`floorEntry` → `get`/`containsKey`** — all three sites, semantics preserved.

**Convention pass** (against `.claude/rules/`): commit format, branch name, base branch, Kotlin naming, test naming/JUnit 4/`CoroutineRule`, no stray `println`, and `lintKotlin` + `lint` in the green `Checks` job all conform. I'd also retract a caveat I raised out-of-band about the `Co-Authored-By` trailer — 61 occurrences in the last 60 commits on `develop`, so it is established practice in this repo and there is nothing to strip.

Two trivial `kdoc-rule.md` nits inline below ("class/function references use `[ClassName]`, not backticks"). Both predate `cb1201f` — I missed them on the first pass. Non-blocking; fold them in if you touch these files again.

A third, `` `createdAt` `` in `CrdtTree.kt`, I'm deliberately leaving alone — it's a property rather than a class or function, so it reads as outside the rule's scope.

PR #363, opened 2026-09-14
release: v0.7.14

thiennguyen-naver:
Release v0.7.14.

## Changes since v0.7.12

- Sync the Android SDK with Yorkie JS SDK v0.7.13 (#359) and v0.7.14 (#360).
- Support `initialRoot` in `Client.attachDocument` and history-exempt local updates via `updateAsync(skipHistory = true)`; make history-generated TreeEdit undo/redo converge on remote replicas (#358).
- Add identity-preserving restore for `JsonText` undo/redo — undo revives the original node IDs instead of inserting copies; adds `RestoreSpan` / `RestoreMode` to `Operation.Edit` (ordinary edits stay wire-identical).
- Add identity-preserving restore for `JsonTree` undo/redo with `TreeRestoreSpan` on `TreeEdit`; fix the tree-merge family (tombstones move with merge, chained merges flatten, style ranges no longer cross a concurrent merge anchor).
- Fix the `splitLevel` root-walk guard, register GC pairs for pieces split off tombstoned nodes, and keep purged text runs in document order on undo after GC.
- Tree undo offsets fail loudly on an out-of-range stale offset (as in JS); Text undo offsets are clamped to the current content length.
- Document the compatibility floor for identity-preserving undo/redo (server ≥ 0.7.14 and peer SDKs) in the README `### Compatibility` section.
- Four tree defects shared with JS SDK v0.7.14 are ported as-is and pinned in `TreeUpstreamDefectPinTest` (upstream-first; see #360 "Known imported defects").
- Update local and CI Yorkie server images to 0.7.14.
- Bump `VERSION_NAME` to 0.7.14.

## Verification

- [x] `./gradlew formatKotlin`
- [x] `./gradlew clean`
- [x] `./gradlew lintKotlin yorkie:testDebugUnitTest`
- [x] `./gradlew yorkie:jacocoDebugTestReport`

## Test plan

- [ ] Run `JsonTextRestoreTest` and `JsonTreeRestoreTest` against a Yorkie 0.7.14 server.
- [ ] Two-client sanity in an example app: concurrent overlapping tree deletes, undo on both — content and node identities converge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added identity-preserving undo/redo for text and tree content, including concurrent edits and garbage-collected content.
  - Added optional document root initialization during attachment.
  - Added updates that can bypass undo history while preserving synchronization and history reconciliation.

- **Bug Fixes**
[...]

PR #364, opened 2026-09-14
chore: sync v0.7.14 back to develop

thiennguyen-naver:
Sync `main` back into `develop` after the v0.7.14 release (#363).

Brings the `VERSION_NAME=0.7.14` bump and the release merge commit onto `develop` so the branches stay consistent with tag `v0.7.14`.


longhoang-lewy (approved):
Approving — routine post-release back-merge, verified clean.

**What I checked**

- **The diff is only the version bump.** `develop...main` touches exactly one line: `VERSION_NAME=0.7.12` → `0.7.14`. No code on `main` is missing from `develop`, so the merge cannot regress any of the 0.7.13/0.7.14/0.7.15 sync work already on `develop` (#361 landed as `7d23d17`).
- **The version is right.** `main` and tag `v0.7.14` agree at `0.7.14`, and the tag points at exactly this PR's head (`9868aa8`). `develop` carrying `0.7.12` while holding newer sync work is the expected state for this repo — the bump happens at release, not at sync — so `0.7.14` is the correct value to land, not `0.7.15`.
- **It matches precedent.** #357 (`v0.7.12`), #354 and #350 are the same `main` → `develop` back-merge, each a single `gradle.properties` line landed as a merge commit.
- CI green on `9868aa8` (Checks, instrumented + unit tests on API 33, Codecov, Pages); `mergeable=CLEAN`.

One note for whoever merges: please use a **merge commit, not squash**. Squashing a back-merge collapses `main`'s history into a new commit on `develop` and leaves the branches without a shared ancestor, which makes the next release's back-merge show phantom conflicts. The three prior back-merges were all merge commits.

PR #365, opened 2026-09-15
release: v0.7.15

thiennguyen-naver:
## Summary
Release v0.7.15 of the Android SDK. Cut from `develop` at `e22af15c`; bumps `VERSION_NAME` 0.7.14 → 0.7.15. Ports yorkie-js-sdk v0.7.15 and pins the companion server image to 0.7.15.

## What's in this release (vs v0.7.14)
- **Tree restore is split-aware for concurrent overlapping undo after GC** (#361, RTCOLLABPLATFORM-755, JS `7b2ab7a4` / yorkie-js-sdk#1315). `CrdtTree.restore` and `retombstone` now isolate a text piece that straddles a restore-span boundary instead of skipping it, so overlapping undo/redo across replicas no longer loses content or diverges text-node segmentation. Restore returns pending GC pairs and drains born-removed split remainders on every path.
- **Typed errors for tree restore preconditions.** `isolateTextRange` and `recreateFromSpan` missing-value cases throw `YorkieException(ErrInvalidArgument)` instead of an unchecked exception.
- **Int counter wrap parity** (JS `bb3c73dc` / yorkie-js-sdk#1312). No code change needed — Kotlin `Int` arithmetic already wraps mod 2^32, congruent with the fixed JS/Go; pinned by ported unit, two-replica, and instrumented tests.
- **Zero-length text span decode kept accepted** for JS/iOS parity (the Android-only rejection added in an earlier review round was reverted).
- **Docker server pin** 0.7.14 → 0.7.15 in `docker/docker-compose*.yml`.
- **Test-only:** presence watched/changed/unwatched test made independent of stream timing (`001874af`); tree restore overlap matrix relayed over the wire; undo-order, interleaved-sync, and GC-pair-ordering exactness pins.

No proto, converter, or wire-format changes. No public API changes.

## Commits
```
722cb347 chore: bump version to 0.7.15
e22af15c Merge pull request #364 from yorkie-team/main
7d23d176 Merge pull request #361 from yorkie-team/feat/RTCOLLABPLATFORM-755-sync-js-v0715
8f2c7709 test: relay the tree restore overlap matrix over the wire and dedupe relations
e9074b5f fix: typed errors for tree restore preconditions, restore zero-length span parity
001874af test: make presence watched/changed/unwatched test independent of stream timing
1198e94f test: harden split-aware tree restore per review 5036530353
7cd52014 test: pin undo-order, interleaved-sync, and GC-pair-ordering exactness for split-aware restore
a91d54af feat: close js sdk v0.7.15 sync-up (RTCOLLABPLATFORM-755)
53d87cdb test: refresh TreeRestoreConvergenceTest class KDoc for spec-006 cases (round-2 LOW-1)
82bd4b10 build: round 1 — 006_sync-js-sdk-v0715
```

## Test plan
- [x] `./gradlew clean lintKotlin yorkie:testDebugUnitTest yorkie:jacocoDebugTestReport` on the release branch — BUILD SUCCESSFUL, 759 tests, 0 failures, 3 skipped (2 upstream it.skip twins + 1 pre-existing).
- [ ] After merge: `VERSION_NAME` on `main` equals `0.7.15` before the GitHub Release is created.
- [ ] `gh release create v0.7.15 --target main` fires `publish.yml` → Maven Central.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved counter synchronization for very large increments, ensuring replicas converge consistently.
  * Improved undo and redo behavior for overlapping text and tree edits, including edits restored after cleanup.
  * Improved handling of zero-length text restoration across compatible clients.
  * Improved presence event synchronization by replacing timing-sensitive waits with reliable event confirmation.

* **Chores**
  * Updated the Yorkie release and Docker images to version 0.7.15.


longhoang-lewy (commented):
## Review — release v0.7.15

Approve-with-follow-ups. The port is faithful and I reproduced the test plan independently; the issues below are process/verification gaps around the release, not defects in the ported logic.

### Verified (not just read)

**Upstream scope is exactly right.** `git log v0.7.14..v0.7.15` on yorkie-js-sdk is precisely 3 commits: the release commit plus `7b2ab7a4` (tree restore) and `bb3c73dc` (counter wrap). Both are addressed; nothing was missed.

**The port is faithful.** `restore` / `retombstone` / `isolateTextRange` in `CrdtTree.kt:1439-1592` are a line-by-line translation of JS `tree.ts`, and they also match the iOS 0.7.15 port (`CRDTTree.swift:2601-2750`) — same 4-tuple shape, same drain-pending-pairs-before-unregister ordering, same `max(length, 1)` clamp. I walked the cursor/`pieceIndex` loop against the mutated-piece aliasing (`pieces` is a snapshot, but `pieceEnd` is read pre-split each iteration) and it is correct.

**The counter "no code change needed" determination is sound.** `int32(int32(a) + int32(b)) ≡ int32(a + b)` mod 2^32, so `value.toInt() + primitiveValue.toInt()` under Kotlin's wrapping `Int` arithmetic is congruent to the fixed JS `bigintToInt32(BigInt(value) + delta)`. The specific bug JS had — truncating the delta but leaving the result an unwrapped JS `number` — cannot occur in Kotlin because the result type *is* `Int`. `1 + 5_000_000_000 → 705032705` checks out.

**Test plan reproduced locally.** `./gradlew lintKotlin yorkie:testDebugUnitTest` → BUILD SUCCESSFUL, **759 tests, 0 failures, 3 skipped** — the claimed numbers exactly. The 3 skips are the 2 JS `it.skip` twins (1:1 name match with `history_tree_concurrent_test.ts:143,186`) plus one pre-existing `CrdtTreeTest` skip.

**Release mechanics.** `origin/main` is fully contained in the PR head (clean cut, no merge surprises); the PR is `develop@e22af15` + the version bump only; bump-on-release-branch matches #363's pattern; `yorkieteam/yorkie:0.7.15` exists on Docker Hub; the `OperationConverter.kt` diff is comment-only, so "no wire changes" holds.

### Findings

**1. CI never runs on this PR.** `check_build_and_test.yml` triggers on `pull_request` to `develop` and `release/**` — not `main`. `gh pr checks 365` returns only `license/cla`. So the commit that `publish.yml` ships to Maven Central reaches `main` with zero automated verification. This is pre-existing for every release PR in this repo; worth adding `main` to the trigger list. My local run substitutes this time.

**2. The instrumented tests in this PR are verified nowhere.** Three of the fifteen changed files are `androidTest`: `JsonCounterWrapTest`, `JsonTreeRestoreTest`, `DocPresenceTest`. `testDebugUnitTest` does not run them, and per (1) neither does CI. That is awkward here specifically because the `DocPresenceTest` change *is* a flake fix for the instrumented suite, and the docker pin moves the server to 0.7.15. Suggest running `connectedDebugAndroidTest` against the 0.7.15 server once before tagging.

**3. `restore`'s `try/finally` drain diverges from JS/iOS, in the worse direction.** On a throw, the `finally` at `CrdtTree.kt:1496` drains `pendingGcPairs`, but `executeRestore` registers them only on normal return — so they are discarded, leaving an unpurgeable tombstone and an undercounted `docSize.gc`. JS and iOS leave them buffered, so the *next* `edit`/`style` drains and registers them: late, but correct. The in-code rationale ("must not leave this op's remainders for the next operation's drain") has it backwards — that carry-over is the recovery, not the bug.

That said, I traced both throw sites and **neither is reachable in production**: `toTreeRestoreSpan` guarantees `value != null` for `isText` spans (and rejects `value.length != length`), and `isolateTextRange`'s precondition is guaranteed by `restore`'s cursor invariant and `retombstone`'s clamp. So this is latent, not live — a follow-up ticket rather than a release blocker. Note that the test `restore drains pending GC pairs even when a later span throws` actively pins the divergent behaviour, so it will need flipping if the fix later goes the JS way.

**4. On the CodeRabbit comments.** The `assertTrue(purged > 0)` suggestion is worth taking, though its stated reasoning is off: `deleteOverlapping` already asserts `garbageLength == 0` *and* `docSize.gc == DataSize(0, 0)` on both replicas. The real gap is narrower — nothing asserts `afterDeletes != initial`, so in principle the whole matrix could pass vacuously if the edits became no-ops. Cheap hardening, not required. The third comment duplicates (3).

### Notes

The test suite is genuinely strong: 6 overlap relations × undo/undo+redo, over-the-wire relay, interleaved and reverse-interleaved undo order, plus operation-level pins for exactly the two invariants that warranted checking (`pendingGcPairs` registered before unregister; live-split overhead reaching `root.acc`).

Reproducing the build locally needed `JAVA_HOME` pointed at Android Studio's JBR, an `sdk.dir` in `local.properties`, and jacoco `0.8.8 → 0.8.13` (0.8.8 rejects Java 21 class files). CI's Java 17 avoids the last one — no change needed here.

longhoang-lewy (approved):

PR #366, opened 2026-09-16
[Android] Sync-up Yorkie JS SDK v0.7.17

thiennguyen-naver:
> **RTCOLLABPLATFORM-768**: [[Android] Sync-up Yorkie JS SDK v0.7.17](https://jira.navercorp.com/browse/RTCOLLABPLATFORM-768)

## Summary
- Ports yorkie-js-sdk v0.7.17 (one upstream fix) into the Android SDK, plus one correction taken from the iOS port's review rounds, plus the server pin. Base `develop` (PR #362 merged), no stacking. Three commits, one per change.
- Fixes `docSize` accounting when a non-empty container is removed: previously the container's own size moved to gc while its descendants' cost stayed in live, so `gc` went negative after collection (`{data:-2, meta:-48}` for `{k:{a:'1'}}`) and the size-limit gate under-counted the document.
- Adds the `SetOperation` undo deregistration Android never had, and keeps a tombstone nested inside an undo-restored container collectable (a JS-literal port would have regressed this).
- No proto, public API, README or version change.

## Changes

**Upstream change list (JS `v0.7.16..v0.7.17`)**

| JS commit | PR | Android |
|-----------|----|---------|
| `611e6e43` Count a removed container's descendants in the GC total | yorkie-js-sdk#1322 (mirror of yorkie#1934) | `fix: count a removed container's descendants in the gc total` |
| `0099bf2c` Archive a task record, index docs/ and scripts/, gate doc links | #1321 | not applicable (JS repo tooling) |
| `dce5ef72` Release v0.7.17 | #1323 | not applicable (version bump; `/release` owns ours) |

**`CrdtRoot`** — new `sizeInGC` ledger (createdAt → charged `DataSize`) and the idempotent `moveSizeToGC`, the only route into `docSize.gc`. `registerRemovedElement` sweeps a removed container's descendants into gc and refunds the removal ticket only on a live-held move. `deregisterElement` is recursive, returns the count, releases the charged amount from gc (else the current size from live); `garbageCollectInternal` is deleted. New `adoptRemovedElement` (moves element + descendants to gc, adds to the GC set, no ticket refund).

**`SetOperation`** — on `UndoRedo`, deregisters the element **registered** under the incoming createdAt before registering the copy (Android had no block at all: the tombstone's createdAt stayed in the GC set forever and a peer-written member stayed registered), then adopts every `removedAt != null` descendant of the restored copy.

**Android determinations (pinned by tests, not transliterated)**
- *Stale twin* — Android's array-remove reverse is `Add + childSetOps`, so a restored child and its tombstoned twin share a createdAt until GC. Deregistering an instance the pair map no longer points at is a no-op. JS never meets this path (its reverse is one `AddOperation` with children inside).
- *Nested tombstone* — recursive deregistration drops a nested tombstone's createdAt from the GC set and `registerElement` re-books its copy into live; JS v0.7.17 has this defect (yorkie-js-sdk#1349 item 1). Ported the iOS fix (`adoptRemovedElement` walk) from yorkie-ios-sdk PR #271's squash `fd15fa3cf6`, which is not on its `feature/0.7.17` branch.

**Kept for parity with JS v0.7.17 (disclosed, pinned, not fixed)** — the `UndoRedo` gate on the deregister block (a peer applying the same undo as `Remote` keeps a stale ledger; #1349 item 2, `NOTE` at the gate, two-replica pin); the born-removed LWW-loser refund (#1349 item 3; after a concurrent conflicting `set` the replicas' `DocSize` differ by one ticket while JSON converges — measured, pre-existing on `develop`, parity unverified against executed JS); the constructor `live.meta` over-credit on snapshot / `deepCopy` (pinned at `gc` equal, `live.meta` + one ticket).

**iOS cross-check (PR #271)** — iOS constants are one ticket lower per member write (`ElementRHT.set` never assigns `movedAt`); Android assigns it like JS, so JS constants apply unchanged. `restoring a container over a diverged tombstone` is quarantined on iOS (RTCOLLABPLATFORM-767, LWW tie); Android compares `positionedAt < executedAt`, so it runs and passes. The identity guard is Android-only.

**Server pin** `yorkieteam/yorkie:0.7.16 → 0.7.17` in `docker/docker-compose.yml` and `docker/docker-compose-ci.yml`.

**Tests** — new `DocumentSizeContainerGcTest` (six JS cases with the JS constants, `gc == built` asserted exactly after removal; two iOS-only single-client cases; the #1349 item-2 parity pin), `CrdtRootTest` +5, `DocumentSizeTest` +1 (constructor-drift pin), instrumented `DocumentSizeGcTest` (two clients, real server 0.7.17). All eight ported cases were RED against the unmodified source; the nested-tombstone case is also RED against the JS-literal port (collects 0, live `{4,192}`). Note: the first commit alone leaves that one case RED by design (the adopt walk lands in the second commit); the PR squash-merges.

## Test plan
- [ ] Two clients attached to the same document against a 0.7.17 server: client A sets `k = {a:'1', b:'2'}`, client B removes `k`; after sync both report equal `getDocSize()` and `{}`; after a few sync rounds `garbageLength == 0` on both and `getDocSize()` equals an empty document's.
- [ ] Single client: set `k = {a:'1', b:'2'}`, remove `k.b`, remove `k`, undo; the document shows `{"k":{"a":"1"}}`, `garbageLength == 1`, and one element is collected on the next GC (`getDocSize()` then equals a fresh `{k:{a:'1'}}`).
- [ ] Undo the removal of `k = [{a:'1'}]` from an array; after GC the restored member is still resolvable and `garbageLength == 0`.

> Items run automatically by CI (lint, unit tests, coverage, instrumented suite) are excluded.


yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
```diff
@@ -307,11 +410,48 @@ internal class CrdtRoot(val rootObject: CrdtObject) {
...
+     * instance still owns the same descendant createdAts in its own member
+     * table. When a later [garbageCollect] purges that old tombstoned
+     * container and walks ITS descendants, this closure would otherwise
+     * delete the live twin's registration by createdAt. Guard (Android-only
+     * divergence forced by this reverse-op shape, not present in JS/iOS): if
+     * the element table's entry for a createdAt no longer points at THIS
+     * instance, a live twin already owns it — no-op, nothing was actually
+     * collected.
+     */
+    @VisibleForTesting
```
ngocann, line 435: `deregisterElement` is no longer test-only — `SetOperation.kt:56` calls it from production code on the `UndoRedo` path. `androidx.annotation.VisibleForTesting` defaults to `otherwise = PRIVATE`, so this annotation now asserts something false about a real internal API and makes every production call a `VisibleForTests` lint violation.

It didn't fail CI (that check is a warning), but it misleads the next reader about what the method is for. `CrdtRoot` is already `internal`, so visibility is scoped correctly without it — drop the annotation, and the `androidx.annotation.VisibleForTesting` import on line 3 with it (this is the only usage in the file, so `lintKotlin` will flag the leftover import).

longhoang-lewy (commented):
## Review — sync-up to JS SDK v0.7.17

**Verdict: approve, with two follow-ups to disclose/pin before merge.** The port is faithful, the three claimed Android determinations all hold up under test, and I could not find a defect in the `CrdtRoot` ledger itself.

### What I verified (not just read)

| Check | Result |
|---|---|
| Full `:yorkie:testDebugUnitTest` on the PR branch | **811 tests, 0 failures, 3 skipped** |
| `DocumentSizeContainerGcTest` / `CrdtRootTest` / `DocumentSizeTest` | 11 / 10 / 11 pass |
| `./gradlew lintKotlin` | clean |
| CI (checks, Codecov, instrumented API 33, CLA) | all green |
| All callers of the 3 changed `CrdtRoot` APIs | enumerated; `deregisterElement` had no prior production caller |
| Undo→redo→undo cycling (probe) | stable, returns exactly to `built`, `gc=(0,0)` — the ordering hazard between the new deregister block and the pre-existing `removed?.let(registerRemovedElement)` is **not** reachable |

I ran two adversarial probes against the PR branch, the base (`2bfb901`), **and** the merged iOS port (yorkie-ios-sdk#271), to separate port defects from inherited ones.

---

### 1. Android-only: `getDocSize()` diverges between replicas after a *remote* undo of an array-member removal, leaking `gc` permanently

Build `{k:[{a:'1'}]}` → sync → `removeAt(0)` → sync → undo → sync → GC. Measured:

| | d1 (undoer) | d2 (peer) |
|---|---|---|
| PR branch, after GC | `gc=(0,0)` ✅ | **`gc=(2,48)`, `garbageLength==0`** — uncollectable forever |
| base `2bfb901` | `gc=(-2,-48)` | `gc=(-2,-48)` (wrong, but **symmetric**) |
| iOS (merged #271) | `gc=(0,0)` | `gc=(0,0)` |

The peer applies the undo as `Remote`, so the new `SetOperation` deregister block — gated on `UndoRedo` for yorkie-js-sdk#1349 item-2 parity — never runs and `sizeInGC[a]` stays charged. Then the new stale-twin guard in `deregisterElement` correctly refuses to touch the live twin, so nothing ever releases that charge.

This isn't a regression (base was worse), but it does break the invariant this PR makes its headline: T3, T4 and the instrumented test all assert `d1.getDocSize() == d2.getDocSize()`. T9 pins the **object** variant of #1349 item 2; the **array** variant behaves differently — a permanent positive `gc` leak rather than a stale count — and is neither pinned nor listed under "Kept for parity (disclosed)".

Suggestion: either drop the `sizeInGC` entry when `registerElement` re-registers a createdAt currently charged to gc, or add the pin + a disclosure line.

### 2. Upstream, not a port defect: `docSize.live` goes **negative** when text is edited inside an already-removed container

`{k:{t:Text("hello")}}` → sync → d1 removes `k` while d2 concurrently deletes the text range → sync:

- **Android PR**: d1 `live=(-10, 0)` — negative `data`, and `meta` below the empty-document floor of 24 — and it stays there after GC; d2 correct at `(0,24)`.
- **iOS (merged #271): identical**, `live=(-10, 0)` vs `(0,24)`, modulo the documented one-ticket object-member constant.
- base `2bfb901`: `gc=(0,-48)`, converged — so this is the same bug relocated from `gc` to `live`.

Cause: `registerRemovedElement` now sweeps the text's full `getDataSize()` (which sums its live nodes) into gc, then `registerGCPair` debits `live` a second time for the same node bytes.

Both KDocs already flag the area, but the wording covers content *added* to a removed Text — not the double-debit on content *removed*, and neither mentions that `live` can go negative or that replicas diverge. **Since iOS reproduces it exactly, this should not block this PR** — it belongs in yorkie-js-sdk#1349 with these numbers, plus a line in this PR's disclosure list.

### 3. The Android-only identity guard is covering a real iOS gap (supports the design call)

On the merged iOS port, after the array-member undo + GC, `findByCreatedAt` on the restored member returns `nil` on **both** replicas — resolvable before GC, gone after. Android's T6 asserts the opposite and passes.

So Divergence 2 isn't just Android-shape bookkeeping; it's strictly better than iOS, and iOS deserves a follow-up ticket.

### Minor

- `deregisterElement`: the `else` branch also runs when `registered == null`, debiting `live` for something no longer in the table. Not reachable today (`garbageCollect` pre-checks the pair, `SetOperation` passes a `findByCreatedAt` result), but the condition reads as "not a stale twin" when it means "not a *different* instance" — worth making the null case an explicit no-op.
- `sizeInGC` entries are released only on the charged branch of `deregisterElement`; the stale-twin no-op leaves them behind permanently (the mechanism behind finding 1, plus unbounded map growth over a long session).

---

The docs, the divergence rationales, and the test-to-AC mapping are unusually good — each of the three "Android determinations" turned out to be defensible under test, which is not the norm for a port this fiddly.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
thiennguyen-naver: Fixed in 8abf9a73: dropped `@VisibleForTesting` and the `androidx.annotation.VisibleForTesting` import (the only usage in the file). `lintKotlin` is clean.

thiennguyen-naver:
Addressed in 8abf9a73 (unit 813/0/3, `lintKotlin` clean). I reproduced all three findings with your exact numbers before changing anything.

**1. Peer gc leak after a remote array-member undo: fixed.** It was a little worse than reported: if d1 then removes the member again, d2's `live` stays at `{2,120}` while d1 goes to `{0,72}`. The old member's charge sat on the createdAt the restored member now owns, so `moveSizeToGC` only topped it up instead of debiting live.

I took your first option, in `registerElement`: when it replaces a registration held by a *different* instance, it releases that charge. That's what the `UndoRedo` deregister already does on the undoing replica. New T10 checks that both replicas have equal `getDocSize()` after GC and again after a second removal.

I tried keying `sizeInGC` by instance identity first. It fixed the array case, but in the object case it only moved the leak from live to gc, because `ElementRht.set` drops the old tombstone node and it's never purged.

As a side effect, the object variant of #1349 item 2 now agrees on sizes too. T9 is narrowed to pin only what's still stale: the leftover createdAt in the GC set (`garbageLength` over-counts, and nothing gets collected). The PR body describes this as an Android divergence.

**2. Negative `live`: disclosed and pinned, not fixed.** T11 pins d1 `live {-10,0}` against d2 `{0,24}`. The `sizeInGC` KDoc now describes the double-debit on *removed* content, including that live can go negative and replicas diverge. It's also in the PR's parity list, and I've posted the numbers on yorkie-js-sdk#1349.

**3. iOS lookup gap:** noted. No Android change, and no iOS ticket from this PR.

**Minors:** both addressed. `deregisterElement` is now a no-op whenever the table doesn't point at this exact instance, including `registered == null`. Stale-twin entries no longer pile up in `sizeInGC`, because the replacement releases them.

yorkie/src/main/kotlin/dev/yorkie/document/operation/SetOperation.kt
```diff
@@ -42,7 +43,36 @@ internal data class SetOperation(
             val copiedValue = value.deepCopy()
             copiedValue.removedAt = null
             val removed = parentObject.set(key, copiedValue, executedAt)
+            if (source == OpSource.UndoRedo) {
```
longhoang-lewy: **Blocker — unbounded `gc` leak on the remote path.**

This gate means a peer applying the same undo as `Remote` never deregisters the old tombstoned container, so the gc charge of any member the undo snapshot does not carry is leaked permanently.

Measured (d1 sets `k={a:'1'}`, sync; d1 removes `k` while d2 writes `k.b='2'`, sync; d1 undoes; sync; GC to completion):
- d1 → `live(2,120)` / `gc(0,0)` — correct, equals a fresh `{k:{a:'1'}}`
- d2 → `live(2,120)` / `gc(2,48)` — `b`'s charge, which no GC round releases

Repeat the shape three times and d2 reaches `gc(6,144)`; remove every key and settle, and d1 reports the empty-document size while d2 still reports `gc(6,144)`. On base `2bfb901` both replicas agreed, so this is introduced here.

`totalDocSize = live + gc` feeds the `maxSizeLimit` gate at `Document.kt:320`, so d2 is eventually refused updates on a document that is actually empty.

Suggested fix: drop the `source == OpSource.UndoRedo` condition entirely. On `Local`, `findByCreatedAt(copiedValue.createdAt)` is a freshly-issued ticket and the lookup is always null, so the block is a no-op there — the gate buys nothing and costs the remote path. That single change also resolves the two findings below.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
```diff
@@ -307,11 +431,45 @@ internal class CrdtRoot(val rootObject: CrdtObject) {
...
+     */
+    fun deregisterElement(element: CrdtElement): Int {
         var count = 0
         val callback = { elem: CrdtElement, _: CrdtContainer? ->
-            deregisterElement(elem)
-            count++
+            val createdAt = elem.createdAt
+            val registered = elementPairMapByCreatedAt[createdAt]
+            if (registered?.element === elem) {
+                val charged = sizeInGC.remove(createdAt)
```
longhoang-lewy: **Blocker — bytes inserted into an already-removed container are stranded in `live` forever.**

`deregisterElement` releases the *frozen* `sizeInGC` amount, but `acc()` keeps adding to `docSize.live` after the sweep has run. So anything a peer inserts into a `CrdtText` / `CrdtTree` that sits inside an already-removed container is charged to `live` and never released.

Measured (d1 sets `k.t = 'hi'`, sync; d1 removes `k` while d2 does `k.t.edit(2,2,'world')`, sync; GC to completion):
- d1 → `live(10,48)` permanently — 10 data + 24 meta above the empty baseline `(0,24)`
- d2 → correctly empty

`totalDocSize` is 58 on d1 vs 24 on d2, so the size gate over-counts on d1 forever. On base `2bfb901` both replicas agreed — both wrong in the same way, but converged — so this is a new divergence.

Path: `registerRemovedElement` records `t`'s pre-insert size into `sizeInGC`; `EditOperation.kt:148` then does `root.acc(result.dataSize)` straight into `live` with no corresponding top-up of the gc charge; `deregisterElement` subtracts only the frozen amount, leaving the difference in `live`.

T11 pins the *removal* direction of this bug; the *insert* direction is unpinned. Either top up `sizeInGC` when `acc()` targets an element whose subtree is already swept, or recompute at deregistration instead of releasing the frozen figure.

longhoang-lewy (commented):
Reviewed by checking out the head in a worktree, running the 34 new JVM tests (all green), then writing throwaway probe tests and running each one **twice** — once against this branch and once with `CrdtRoot.kt` / `SetOperation.kt` reverted to the merge-base `2bfb901`. Every number below is measured, and the base-vs-head comparison is what separates a regression this PR introduces from the pre-existing drift the description already discloses.

The core `docSize` fix is sound and well covered on the single-replica path — the six JS cases, the exact `gc == built` assertions and the RED-against-base evidence are exactly the right shape. The problem is on the **remote** path.

### The description's central claim does not hold

> sizes agree since `registerElement` releases a replaced instance's gc charge

This is false whenever the peer grew a member that the undo snapshot does not carry. d1 sets `k={a:'1'}`, sync; d1 removes `k` while d2 concurrently writes `k.b='2'`, sync; d1 undoes. JSON converges on both, but d2's ledger keeps `b`'s charge in `gc` **permanently** — no GC round ever releases it:

| | d1 | d2 |
|---|---|---|
| after undo + sync + GC to completion | `live(2,120)` / `gc(0,0)` | `live(2,120)` / `gc(2,48)` |
| same shape ×3 | `gc(0,0)` | `gc(6,144)` |
| then remove every key and settle | empty-document size | still `gc(6,144)` |

On `2bfb901` both replicas agreed. `totalDocSize = live + gc` gates local updates (`Document.kt:320`, `ErrDocumentSizeExceedsLimit`) and `garbageCollect` runs automatically inside `applyChangePack` (`Document.kt:611`), so this is reachable in normal operation: a peer eventually refused writes on a document that is actually empty.

### One root cause covers three of the regressions

The `source == OpSource.UndoRedo` gate in `SetOperation`. On `Local` it is provably inert — `findByCreatedAt(copiedValue.createdAt)` is a fresh ticket, so the lookup is always null. It changes behaviour **only** on `Remote`, which is precisely where the bugs are. Dropping the gate fixes the leak above, the stale `gcElementSetByCreatedAt` entry, and the nested-tombstone divergence together.

### The test that would have caught it

T5 builds this exact scenario and then asserts only d1 — the one two-replica case in the file that checks a single side, and the omitted side is where the leak lives.

---

Inline comments below: 6 correctness findings on `main` (the first two are the blockers), 1 on `garbageCollect`'s missing identity guard, 3 on the instrumented test helpers, 3 on test quality, 1 perf/tidy, 1 convention. Happy to talk through any of them.

yorkie/src/main/kotlin/dev/yorkie/document/operation/SetOperation.kt
```diff
@@ -42,7 +43,36 @@ internal data class SetOperation(
...
+                // Deregisters the
+                // REGISTERED element under the incoming createdAt (the
+                // tombstone being restored, or a member a peer grew on it),
+                // never the incoming copy — copiedValue has not been
+                // registered yet, so findByCreatedAt only ever returns the
+                // previously-registered element here.
+                root.findByCreatedAt(copiedValue.createdAt)?.let(root::deregisterElement)
+            }
             root.registerElement(copiedValue, parentObject)
+            if (source == OpSource.UndoRedo && copiedValue is CrdtContainer) {
```
longhoang-lewy: **Same `UndoRedo` gate — the adopt walk is skipped on the remote path, so the peer books the restored container's nested tombstone into `live` instead of `gc`.**

Measured (d1 sets `k={a:'1',b:'2'}`, sync; removes `k.b`, sync; removes `k`, sync; undoes; sync), immediately after the undo:
- d1 → `live(2,120)` / `gc(2,72)`, `garbageLength` 1
- d2 → `live(4,192)` / `gc(0,0)`, `garbageLength` 3 — the nested `b` tombstone sits in d2's `live`

They converge only at the next `garbageCollect`. On base `2bfb901` the two replicas agreed throughout.

`Document.getDocSize()` is public API, and the divergence window lasts until the min-synced version vector advances — for an offline or idle peer that is arbitrarily long. Removing the gate (see the comment above) fixes this too.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
```diff
@@ -129,36 +151,138 @@ internal class CrdtRoot(val rootObject: CrdtObject) {
...
+     * hitting the new registration. This is what `SetOperation`'s
+     * `UndoRedo` deregister does on the undoing replica; doing it here
+     * makes a peer receiving the same undo as `Remote` agree on sizes
+     * (Android-only divergence from yorkie-js-sdk#1349 item 2 — the stale
+     * [gcElementSetByCreatedAt] entry is left as upstream leaves it).
+     */
+    private fun registerOne(element: CrdtElement, parent: CrdtContainer?) {
+        val createdAt = element.createdAt
+        val prev = elementPairMapByCreatedAt.put(createdAt, CrdtElementPair(element, parent))
+        if (prev != null && prev.element !== element) {
```
longhoang-lewy: **`garbageLength` can never reach 0 again on the affected replica.**

This branch releases the replaced instance's gc charge but leaves its `gcElementSetByCreatedAt` entry behind, so the set permanently over-counts and `garbageCollect` returns 0 for that entry forever — `elementPairMapByCreatedAt[createdAt]` now points at the live restored copy, whose `removedAt` is null, so the filter at line 400 skips it every round.

Measured on the same remote-undo shape: d2's `garbageLength` reads 2, then 4, then 6 across three rounds while d1 stays at 0, and `d2.garbageCollect(maxVector)` returns 0 every time.

Consequences beyond the counter:
- any caller looping `while (garbageLength > 0) garbageCollect(...)` never terminates
- `microbenchmark/DocumentBenchmark.kt:242` and `:274` assert `garbageLength == 0` outright
- `getGarbageElementSetSize()` (line 360) re-walks that now-*live* subtree on every read of the public `Document.garbageLength`

Fix: `gcElementSetByCreatedAt.remove(createdAt)` inside the release branch — or drop the `UndoRedo` gate so the deregister runs and does it for you.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
```diff
@@ -129,36 +151,138 @@ internal class CrdtRoot(val rootObject: CrdtObject) {
...
+            sizeInGC.remove(createdAt)?.let { charged ->
+                docSize = docSize.copy(gc = subDataSize(docSize.gc, charged))
+            }
+        }
+        docSize = docSize.copy(live = addDataSizes(docSize.live, element.getDataSize()))
```
longhoang-lewy: **The `live` / `gc` split now diverges with operation-application order.**

`registerOne` books an element added into an already-removed container straight into `docSize.live`, whereas a replica that applies the same two operations in the opposite order sweeps it into `gc`.

Measured (d1 sets `k={a:'1'}`, sync; d1 removes `k` while d2 writes `k.c='3'`; sync):
- d1 → `live(2,72)` / `gc(2,120)`
- d2 → `live(0,24)` / `gc(4,168)`

Totals match (196 each), so the size gate is unaffected — but `Document.getDocSize()` is public API, and equality of that value across replicas is exactly what the new instrumented `DocumentSizeGcTest` and unit cases T3/T4 assert. On base `2bfb901` both replicas reported identical values.

`registerOne` has no notion of "my parent is already removed"; the recursive sweep this PR adds is what creates the asymmetry. Lower severity than the two blockers since it self-corrects at the next GC, but worth a decision rather than leaving it implicit.

yorkie/src/main/kotlin/dev/yorkie/document/operation/SetOperation.kt
```diff
@@ -42,7 +43,36 @@ internal data class SetOperation(
...
+                // Divergence 3 (yorkie-js-sdk#1349 item 1, iOS fd15fa3cf6): a
+                // tombstone nested inside the restored container must stay
+                // collectable. Android's object-remove reverse deep-copies
+                // the whole tombstoned subtree (ElementRht.deepCopy keeps
+                // tombstones), so a nested member can still carry removedAt
+                // here even though copiedValue.removedAt was just cleared
+                // above. Adopt every such descendant into gc directly — the
+                // restored container itself is untouched.
+                copiedValue.getDescendants { elem, _ ->
```
longhoang-lewy: **The fix is scoped to `SetOperation` only, and the walk only visits `CrdtElement` descendants.**

Two gaps:

**1. `AddOperation` restores tombstone-bearing subtrees with neither the deregister nor the adopt walk.** `RemoveOperation.stripChildren` (`RemoveOperation.kt:100`) strips members only for `CrdtObject`, and `childSetOps` returns empty for anything that is not a `CrdtObject` — so undoing `removeAt` on an array *of arrays* reverses as a single `AddOperation` carrying the whole tombstone-bearing subtree into `AddOperation.kt:39`, which has neither block. Measured after such an undo: `live(4,168)` / `gc(0,48)` where a freshly built equivalent is `live(2,120)` / `gc(2,72)`. Self-corrects at the next GC.

**2. `copiedValue.getDescendants` only yields `CrdtElement`s.** Tombstoned `RgaTreeSplit` nodes and `CrdtTreeNode`s inside a restored text or tree get no `registerGCPair` — only `CrdtRoot.init` registers those — leaving them permanently uncollectable and invisible to `garbageLength`. This one does not self-correct.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
```diff
@@ -275,7 +399,7 @@ internal class CrdtRoot(val rootObject: CrdtObject) {
             val removedAt = pair.element.removedAt
             if (removedAt != null && minSyncedVersionVector.afterOrEqual(removedAt)) {
                 pair.parent?.purge(pair.element)
```
longhoang-lewy: **`purge` is called with no identity guard, one line before `deregisterElement` applies one.**

`ElementRht.purge` (`ElementRht.kt:85-97`) and `RgaTreeList.purge` (`RgaTreeList.kt:326-334`) both resolve `elementMapByCreatedAt[element.createdAt]` and throw `YorkieException(ErrInvalidArgument)` on a miss — neither compares instance identity.

The premise of this PR is that a `createdAt` can be claimed by a *different* instance than the one being collected; that is exactly why `deregisterElement` gained `registered?.element === elem`. In that state `purge` either detaches the **live** twin's node from the parent, or throws — and the throw is uncaught inside the `gcElementSetByCreatedAt.toSet().forEach` loop, aborting the sweep with `elementPairMapByCreatedAt`, `gcElementSetByCreatedAt`, `sizeInGC` and `docSize` half-updated and no recovery path.

I could not construct a reproducer on this branch: the two maps happen to stay in sync because `ElementRht.set` overwrites `nodeMapByCreatedAt` in the same call that leads to `registerOne`. But nothing enforces that, and the guard you added two lines down is an admission that it is not guaranteed. Worth applying the same `===` check here for symmetry.

yorkie/src/test/kotlin/dev/yorkie/document/DocumentSizeContainerGcTest.kt
```diff
@@ -0,0 +1,473 @@
...
+    /**
+     * RTCOLLABPLATFORM-767 (a tie between a peer's write and an undo restore
+     * leaking the peer's member) is iOS-only: iOS `ElementRHT.set` compares
+     * `createdAt` vs `createdAt` and can tie on an undo restore, while
+     * Android `ElementRht.set` compares `getPositionedAt() < executedAt`
+     * (`ElementRht.kt:31`), so the tie cannot occur here. This case runs
+     * un-quarantined on Android.
+     */
+    @Test
+    fun `restoring a container over a diverged tombstone`() = runTest {
```
longhoang-lewy: **T5 is the only two-replica case in this file that asserts a single replica — and the omitted side is where the leak lives.**

The case sets up the concurrent `d1.remove("k")` / `d2["k"]["b"]="2"` shape and calls `d1.history.undoAsync()`, then asserts only `d1.toJson()`, `d1.getDocSize()` and `d1.garbageLength`. There is no `crossSync(d1, d2)` after the undo and no d2 assertion anywhere. Every sibling case pins both sides (T3:157-158, T4:187-188, T9:391-392, T10:427-428).

Adding `crossSync(d1, d2)` plus `assertEquals(d1.getDocSize(), d2.getDocSize())` makes this case go RED and pins the blocker.

Separately: the KDoc argues the LWW-tie quarantine seen on the iOS port cannot occur here because Android compares `positionedAt < executedAt`. That is plausible, but with no d2 assertion the test provides no evidence for the claim it makes.

yorkie/src/test/kotlin/dev/yorkie/document/DocumentSizeContainerGcTest.kt
```diff
@@ -0,0 +1,473 @@
...
+        d1.updateAsync { root, _ -> root.remove("k") }.await()
+        d2.updateAsync { root, _ ->
+            root.getAs<JsonObject>("k").getAs<JsonText>("t").edit(0, 5, "")
+        }.await()
+        crossSync(d1, d2)
+
+        val vector = maxVectorOf(listOf(actor1, actor2))
+        d1.garbageCollect(vector)
+        d2.garbageCollect(vector)
+        assertEquals(DocSize(live = DataSize(-10, 0), gc = DataSize(0, 0)), d1.getDocSize())
```
longhoang-lewy: **T11 hard-codes a negative, replica-divergent `DocSize` as the expected value, with no `@Ignore` and no tracking ticket.**

```
assertEquals(DocSize(live = DataSize(-10, 0), gc = DataSize(0, 0)), d1.getDocSize())
assertEquals(emptyDocSize, d2.getDocSize())
```

As written this certifies three things as correct: that `live.data` may be negative, that `live.meta` may be 0 where an empty document's is 24, and that two replicas may permanently disagree.

It also does not reproduce the upstream defect unchanged — it relocates it. Base `2bfb901` produces `live(0,72)` / `gc(0,-48)`, negative in `gc`; this branch produces `live(-10,0)` / `gc(0,0)`, negative in `live` — which is the component feeding `totalDocSize` and the `maxSizeLimit` gate. That is a meaningful change in blast radius and reads as an unchanged parity pin.

When upstream fixes the linked js-sdk issue, this test goes RED displaying a *correct* `DocSize` as "actual", and the next engineer reads a genuine fix as a regression. Please either `@Ignore` it with the ticket reference, or assert the invariant you actually mean (e.g. "the two replicas disagree, pending #1349") rather than the literal corrupt tuple.

yorkie/src/androidTest/kotlin/dev/yorkie/core/TestUtils.kt
```diff
@@ -220,8 +236,8 @@ suspend fun awaitWatchConnected(
     val connected = launch(start = CoroutineStart.UNDISPATCHED) {
```
longhoang-lewy: **`withTimeout` inside `launch`, followed by `join()`, silently swallows the timeout.**

`TimeoutCancellationException` is a `CancellationException`, so the `UNDISPATCHED` child completes as *cancelled* rather than *failed*. That does not cancel the parent `coroutineScope`, and `Job.join()` never rethrows. The helper therefore returns normally after `GENERAL_TIMEOUT` and the peer attaches anyway — turning the guard into a 15-second no-op plus the original flake, now slower and harder to diagnose.

This shape pre-dates the PR, but the PR promotes it from one call site to the main `withTwoClientsAndDocuments` path used by every Realtime instrumented test, so it is worth fixing here.

Use `connected.await()` on an `async`, or hoist the `withTimeout` outside the `launch` so the exception propagates.

yorkie/src/androidTest/kotlin/dev/yorkie/core/TestUtils.kt
```diff
@@ -220,8 +236,8 @@ suspend fun awaitWatchConnected(
...
         withTimeout(timeoutMs) {
             document.events
-                .filterIsInstance<Document.Event.StreamConnectionChanged>()
-                .first { it == Document.Event.StreamConnectionChanged.Connected }
+                .filterIsInstance<Document.Event.PresenceChanged.MyPresence.Initialized>()
```
longhoang-lewy: **`MyPresence.Initialized` is also published on every watch-stream failure, so a disconnect satisfies this wait.**

`Client.kt:812` is `document.publishEvent(Initialized(document.presences.value))` inside `onWatchDocumentStreamCanceled`, which runs on every stream failure, close and timeout. The KDoc here says the event means "the server's acknowledgement of the watch stream" — it does not distinguish the two.

On exactly the slow/loaded runner this change targets: a transient watch-stream drop during attach publishes `Initialized`, `.first()` returns, client2 attaches before the server has re-registered the watcher, and the `Others.Watched` assertion fails as before — with the new code now masking the cause.

It is also not a direct signal: `Document.publishEvent` diverts every `PresenceChanged` into `pendingPresenceEvents` (`Document.kt:1027`), drained only when the `presences` StateFlow emits and `presenceEventReadyToBePublished` passes. Worth waiting on something that can only mean *connected*.

yorkie/src/androidTest/kotlin/dev/yorkie/core/TestUtils.kt
```diff
@@ -109,11 +109,21 @@ fun withTwoClientsAndDocuments(
...
+                    document1,
+                    syncMode = syncMode,
+                    initialPresence = presences.first,
+                ).await()
+            }
+            // In realtime mode, document1's watch stream must be registered
+            // server-side before client2 attaches, or client2 arrives in the
+            // Initialization frame instead of as Others.Watched.
+            if (syncMode == Client.SyncMode.Realtime) {
+                awaitWatchConnected(document1) { attach1() }
```
longhoang-lewy: **The guard misses two sync modes and the three-client helper.**

`Client.kt:1190` opens the watch loop for every mode except `Manual` and `Polling`:

```kotlin
if (syncMode != SyncMode.Manual && syncMode != SyncMode.Polling) runWatchLoop(documentKey)
```

The gate here is `if (syncMode == Client.SyncMode.Realtime)`, so `RealtimePushOnly` and `RealtimeSyncOff` take the bare `attach1()` branch and keep the exact race this is fixing. `syncMode !in setOf(Manual, Polling)` would match the production condition.

Separately, `withThreeClientsAndDocuments` (line 150) still does three bare sequential `client.attachDocument(...).await()` calls with no `awaitWatchConnected` at any of them, so every three-client Realtime test retains the race — including the `DocPresenceTest` cases this diff touches, which had to wrap c1's attach by hand. Pushing the guard into the shared helper would remove those ad-hoc wrappers.

yorkie/src/test/kotlin/dev/yorkie/document/crdt/CrdtRootTest.kt
```diff
@@ -196,4 +203,203 @@ class CrdtRootTest {
...
+
+        // when
+        val released = root.deregisterElement(stale)
+
+        // then: nothing changes — the live twin's registration survives.
+        assertEquals(0, released)
+        assertEquals(docSizeBefore, root.docSize)
+        assertEquals(elementMapSizeBefore, root.elementMapSize)
+        assertEquals(garbageLengthBefore, root.garbageLength)
+        assertEquals(replacement, root.findByCreatedAt(replacement.createdAt))
```
longhoang-lewy: **Identity guard pinned with a structural assertion.**

`deregisterElement`'s guard is referential — `registered?.element === elem` — but this asserts `assertEquals(replacement, root.findByCreatedAt(replacement.createdAt))`. `CrdtPrimitive` is a `data class` whose generated equality covers `_value` / `createdAt` / `movedAt` / `removedAt`, so a structurally equal *different* instance satisfies it.

It discriminates today only because the twins were given different values (`'v1'` / `'v2'`). Model the real array-undo shape — a `deepCopy` with identical fields — and the assertion becomes vacuous, which is the case the guard actually exists for. `assertSame` is the matching assertion.

Unrelated, in the same vein: the new AC8 case at `DocumentSizeTest.kt:771` passes unchanged against base `2bfb901` (I ran this PR's test files at the merge-base — `DocumentSizeContainerGcTest` 13/13 RED, `DocumentSizeTest` 11/11 GREEN), because `CrdtRoot.deepCopy()` replays the same `registerRemovedElement` the live root used, so `clone.gc` tracks `live.gc` under any implementation of the sweep. It pins nothing this PR changes.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
```diff
@@ -129,36 +151,138 @@ internal class CrdtRoot(val rootObject: CrdtObject) {
...
+        val moved = moveSizeToGC(element)
         if (element is CrdtContainer) {
-            element.getDescendants { elem, par ->
-                elementPairMapByCreatedAt[elem.createdAt] = CrdtElementPair(elem, par)
-
-                docSize = docSize.copy(
-                    live = addDataSizes(docSize.live, elem.getDataSize()),
-                )
-
+            element.getDescendants { elem, _ ->
```
longhoang-lewy: **Perf + tidy (non-blocking).**

`registerRemovedElement` went from O(1) to a full subtree sweep whose per-node cost is O(content) for `CrdtText` / `CrdtTree` — `CrdtText.getDataSize()` (`CrdtText.kt:346`) iterates the entire `rgaTreeSplit`, `CrdtTree.getDataSize()` (`CrdtTree.kt:1959`) walks the whole tree. Removing a container holding a large document text now costs a full content scan, executed once for the live root and once for the clone (`Document.kt:963-964`). Removing nested containers outermost-first re-sweeps the same subtree once per level.

Three smaller things:
- `moveSizeToGC`'s top-up branch (line 204) allocates two `DataSize` objects plus a `DocSize` per descendant even when the delta is zero — the common case for a re-sweep. `if (size == charged) return false` costs nothing.
- Line 208 builds `DataSize(size.data - charged.data, size.meta - charged.meta)` by hand, although `subDataSize(size, charged)` is imported and used at six other sites in this file.
- `adoptRemovedElement` (line 278) is `registerRemovedElement` (line 244) minus the ticket refund. The "composes cleanly with a later JS sync" rationale argues for one function with a boolean parameter, not two bodies to keep in step.

yorkie/src/main/kotlin/dev/yorkie/document/crdt/CrdtRoot.kt
```diff
@@ -44,6 +43,29 @@ internal class CrdtRoot(val rootObject: CrdtObject) {
      */
     private val gcPairMap: MutableMap<GCChild, GCPair<*>> = IdentityHashMap()
 
+    /**
```
longhoang-lewy: **Convention (non-blocking).**

`.claude/rules/kdoc-rule.md` — tracked in this repo and scoped to Kotlin sources — requires class and symbol references to use `[Brackets]` rather than backticks. The new blocks use backticked `acc`, `gcOnlySize`, `611e6e43`, `fd15fa3cf6`, `UndoRedo`, `Add + childSetOps`, `RemoveOperation.childSetOps` throughout `CrdtRoot.kt:46-67`, `:163-175`, `:434-456` and `SetOperation.kt:47-74`.

More substantively: the SDK rules say the document model must stay consistent across the four SDKs, and that a change to a shared data shape gets chalked in the TDD first and consumed by the clients only once agreed. This PR ships three Android-only determinations — `adoptRemovedElement`, the `deregisterElement` identity guard, and the `registerOne` gc-charge release — that change observable `DocSize` semantics relative to JS v0.7.17 and the iOS port, documented only in Kotlin comments and the PR body. Given the js-sdk issue is already open, folding these into it (or a contract note) would keep the four ports from drifting further apart.

